PR #8, opened 2017-12-30
Add BW Roamer guide

zaksabeast:
# Guide Details

## I want to (select one)

- [x] Add a guide

## Games affected

- Black, White

## Type of RNG

Roamers

## Language

English

## Notes

Guide donated by /u/Zapdos44

### DIFF
--- a/guides/bw/en/roamers (emu).md
+++ b/guides/bw/en/roamers (emu).md
@@ -1,0 +1,105 @@
+# Gen 5 Roamers (Emulator)
+_Original Guide by /u/Zapdos44_
+
+## Preparation
+
+### Make sure you have the following
+- RNGReporter
+- Generation 5 Lua Scripts
+- Generation 5 RNG Experience
+- Generation 5 TID/SID
+- DeSmuMe dev build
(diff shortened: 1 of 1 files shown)


RichardPaulAstley:
Just an tip (not sure if you can add this to the guide) : You can trade the roamer from another game. If you're aiming for a Shiny, you don't need to find it and you can just check your Pokedex to see if a Shiny form is registered. If so : win.

RichardPaulAstley:
Always good for people that do retail in the future, since the guide has no indication on retail or emu in the end (could be adapted for both)

zaksabeast:
The guide's name and first line both mention it's for emulators, so Fish's idea is great.

However I agree with Vlad that extra tips would be good for those who choose to adapt this for retail.

RichardPaulAstley:
Yep but I guess EVERY method / tip is really welcome in that kind of case haha

zaksabeast:
I believe that info was actually in the guide.

Admiral-Fish:
On retail you can use ivs + characteristic to narrow down pid

PR #45, opened 2018-03-06
Added timeline leap method to Gen 7 timeline guide

ShinySylveon04:
# Guide Details

## I want to (select one)

- [ ] Add a guide
- [x] Modify an existing guide

## Games affected

- SM/USUM

## Type of RNG

- Gen 7 Timeline Leap

## Language

- English


wwwwwwzx:
Nice guide.
Some notes here:
- The time interval between step 8 and step 9 is around 10 seconds for wc7, menu method should be less than 3 secs, user should be prepared
- Sometimes the original IVs spread is not on the new timeline, even you can hit the exactly the same frame number. Since NPCs' blink status will affect this
- To solve the above problem: (1) Click "No" when prompted to check if other leap frames can result in the desired spread
(2) Change your target frame by +-1 ~ 4 to see if the desired spread is on other frame number. 
- The spread checks should be done before you unpause from the safe frame, leap from old timeline and hit the target frame.

PR #65, opened 2018-08-12
Add USUM SOS Guide

ShinySylveon04:
# Guide Details

## I want to (select one)

- [x] Add a guide

## Games affected

- USUM

## Type of RNG

- SOS RNG

## Language

- English


guides/usum/en/sos.md
```diff
@@ -0,0 +1,238 @@
...
+- [Encounter slots](https://gist.github.com/SciresM/a539739085e24af55dffdf443cb70eb2)
+  - Can Ctrl + F for Pokemon wanting to SOS.
+
+## Regarding SOS RNG
+
+SOS RNG consists of calls to two different processes:
+
+- Main RNG - IV spread, nature, gender and PID (shininess).
+- G7 SFMT (32bit) (the SOS RNG) - Encounter slot, call success, level.
+  - This is the RNG that will allow you to RNG that 1% Salamence.
```
RichardPaulAstley: I'd switch both, in order to give an idea what you'll have to do in order so : 
- G7 SFMT
- Main RNG

in order of the RNG process

guides/usum/en/sos.md
```diff
@@ -0,0 +1,238 @@
...
+	- HP should be set to what the current caller's HP level is.
+	- Check `Adrenaline Orb` if you have used an Adrenaline Orb.
+	- If there is weather effects such as rain check the `Weather` box.
+	- If your Pokemon currently out has the Intimidate, Unnerve, or Pressure ability check the `Intimidate` box.
+	- If the ally Pokemon that was called was hit with a super-effective attack on the first turn it appeared, then check the `Super Effective` box.
+
+```
+Note: The "Same Caller" box should be checked as long as the Pokemon that does the SOS call is the same as last turn. If it faints and the other Pokemon calls for help, then uncheck the box.
+```	
+
```
RichardPaulAstley: I'd add a note to NOT setup during the first turn, but once the Pokemon is low life + Adrenaline orb is used

guides/usum/en/sos.md
```diff
@@ -0,0 +1,238 @@
...
+5. Once your current SOS frame will trigger an encounter on the next SOS call (has `O` for first two values) you will want to use it to calibrate for the main RNG delay.
+	- You can advance SOS frames by using an Adrenaline Orb (this has no effect if already used one and item will not be consumed) or by knocking out the non-caller Pokemon.
+
+```
+Note: Pay attention to the SOS Current Seed shown in PCalc and compare to what it should be in the Misc. RNG Tool window (Rand#) for the SOS frame you are on.
+
+Sometimes SOS frames do not advance as predicted and will throw off results if not corrected. If the SOS Current Seed does not match, then find the SOS frame in Misc. RNG Tool that matches the SOS Current Seed that PCalc shows. Adjust the SOS frame as needed in both Misc. RNG Tool and the main window.
+```
+
```
RichardPaulAstley: A note for Fish RNG not being predictable

guides/usum/en/sos.md
```diff
@@ -0,0 +1,238 @@
...
+![](https://imgur.com/m6kHHrr.png)
+
+## If you did not obtain the Pokemon you wanted
+
+1. Double check that all info is correct on the two windows. Sometimes the Initial Seed is not typed in correctly, or SOS Frame is not correct.
+
+2. Restart the guide from the beginning and make sure to follow _all_ instructions given. Not getting the correct Pokemon is usually a result of user error.
+
+3. Make sure you are using the correct delay based on the type of SOS call.
```
RichardPaulAstley: Could just be the delay being off ? The 3. can be read as one delay per method. And not "many delay" for each method.

PR #73, opened 2018-09-21
Add XY Trainer ID, Secret ID, TSV RNG Guide

Subject21J:
# Guide Details

## I want to (select one)

- [x] Add a guide

## Games affected

- List all affected games here: X/Y

## Type of RNG

Describe the RNG - Trainer ID RNG

## Language

Please list the language your guide is written in.
- English

## Notes

I know there are two other pull requests for this type of guide, I am not sure which one the team will add.


guides/xy/en/Trainer ID RNg.md
```diff
@@ -0,0 +1,94 @@
...
+
+This is the final screen to change character selection.
+
+## Tiny Seeds
+In order to RNG our TID/SID/TSV we must look at the tiny seeds instead of frames passed. Each time the tiny seeds advance to the next set of tiny seeds a TID/SID/TSV frame advancement happens. 
+In X/Y the tiny seeds will constantly advance throughout the introduction. They will advance at a random pace but will be in a consistent order that 3DSRNGTool can predict.
+Due to this constant advancement it is possible to miss your wanted tiny seeds. To minimize the possibility of missing the required tiny seed, it useful to use PCalc's pause and single frame advancement options. 
+
+```
+Note: Trainer Data generation happens some time during a cut scene and the last user input. This causes a slight delay that will need to be adjusted for. It's cover later in the guide.
```
zaksabeast: Please change
```
It's cover later in the guide
```
to
```
It's covered later in the guide
```

guides/xy/en/Trainer ID RNg.md
```diff
@@ -0,0 +1,94 @@
...
+	- There is a delay that needs to be accounted for here, it is covered in the next section.
+- Wait until you can control your character to check your trainer data. 
+
+![](https://i.imgur.com/aqh4RL4.png) 
+
+```Note: PCalc can display TSV information by pressing `Start+Right` and then `Select+Left` ```
+	
+
+## Cut Scene Delay
+As stated earlier the actual trainer data is generated during a cut scene while the tiny seeds are still advancing. This causes a delay, about 5-10 frames from the target. This delay is unique for every game and system. So stopping several frames before the actual seed may be required to reach the right TID/SID/TSV. To find the delay for your game subtract the target frame from the real frame gotten, this is how many frames you have to stop before the real target. Do multiple dry runs on low seed advancements to find a common delay. Once there is a constant delay then use that for actual TID/SID/TSV RNG. In example given here the delay was 7 frames as shown below.
```
zaksabeast: I believe this should be broken up into two paragraphs:
```
As stated earlier the actual trainer data is generated during a cut scene while the tiny seeds are still advancing. This causes a delay, about 5-10 frames from the target. This delay is unique for every game and system. Stopping several frames before the actual seed may be required to reach the right TID/SID/TSV.

To find the delay for your game subtract the target frame from the real frame gotten, this is how many frames you have to stop before the real target. Do multiple dry runs on low seed advancements to find a common delay. Once there is a constant delay then use that for actual TID/SID/TSV RNG. In example given here the delay was 7 frames as shown below.
```

What are other people's thoughts?

Subject21J: According to wwwwwwzx  Tiny Timeline Tool can be used to avoid the weird delay. I will update the guide later once I have time to see how to use TTT is used for TID RNG so this section will have to be removed.

guides/xy/en/XY Trainer ID, Secret ID, and TSV RNG Guide.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Game Setup
+
+```
+Note: If you are using a N3DS or N2DSXL you can speed up the game when the PCalc overlay is active by enabling L2 cache + 804 MHz in the Rosalina menu options (Select + L + Down to bring up menu).
+```
+
+1. Load the game and connect to NTR Helper in 3DSRNGTool.
+
+   - Instructions for doing so can be found [here].
```
ShinySylveon04: This needs to have a link to the NTR Helper guide once that is added to the site.

ShinySylveon04:
I removed the section about finding your delay since that will be replaced by a section for using Tiny Timeline Tool to predict the delay. I'm not sure how that works, but once I figure it out and add it to the guide, this PR will be ready for review.

PR #94, opened 2019-02-20
Add XY Friend Safari Guide

ShinySylveon04:
# Guide Details

## I want to (select one)

- [x] Add a guide

## Games affected

- XY

## Type of RNG

- Friend Safari RNG

## Language

- English


guides/xy/en/Friend Safari RNG Guide.md
```diff
@@ -0,0 +1,104 @@
+# XY Friend Safari RNG Guide
+
+## Tools
+
+    - [3DSRNGTool](https://github.com/wwwwwwzx/3DSRNGTool/releases)
+
```
zaksabeast: Please remove this extra line.

guides/xy/en/Friend Safari RNG Guide.md
```diff
@@ -0,0 +1,104 @@
...
+    - [PCalc for XY](https://pokemonrng.com/downloads/pcalc/pcalc-xy)
+
+## Required Reading
+
+    - [TinyMT Timeline](https://github.com/wwwwwwzx/3DSRNGTool/wiki/Gen6-TinyMT-Timeline-Calibration)
+    - [NTR Helper Usage](https://github.com/wwwwwwzx/3DSRNGTool/wiki/NTR-Helper-Usage)
+
+## Set up
```
zaksabeast: Please add `Step 1:` here to be consistent with our other guides.

guides/xy/en/Friend Safari RNG Guide.md
```diff
@@ -0,0 +1,104 @@
...
+
+3. Walk to the left side of the area outside of the grass and save.
+
+   - You can now either choose to do a test run, or reset game to do the reseeding method mentioned further in the guide to find a target frame.
+
+```
+Note: This guide assumes you already have basic knowledge of how to do Gen 6 RNG. Otherwise it is recommended to first familiarize yourself with Gen 6 RNG mechanics by RNGing something simple like eggs.
+```
+
+## Explanation of TinyMT frames within Tiny Timeline Tool
```
zaksabeast: Let's move this above `Step 1` since it's not really a step of instructions, but more of required reading before beginning.

Perhaps we should put this in the `Required Reading` section?

guides/xy/en/Friend Safari RNG Guide.md
```diff
@@ -0,0 +1,104 @@
...
+    - **Enctr?**: Determines if there will be an encounter or not if the character turns or walks into the grass during those TinyMT seeds.
+        - Encounter of less than 5 is best since this will guarantee an encounter every time.
+    - **Sync?**: Determines if nature will be Synchronized or not.
+        - `X` = No `O` = Yes
+    - **Slot**: Determines which slot of Friend Safari will be chosen for the Pokemon that appears.
+        - Order is same order as shown in Friend Safari Pokemon screen from left to right before entering the specific safari.
+    - **HA**: Determines if Pokemon will have HA or not.
+        - `X` = No `O` = Yes
+
+## Manipulating TinyMT
```
zaksabeast: Please add `Step 2:` here to be consistent with our other guides.

guides/xy/en/Friend Safari RNG Guide.md
```diff
@@ -0,0 +1,104 @@
...
+
+The key to successfully causing an encounter on your target frame is manipulating the TinyMT using the method below. It is suggested to practice manipulating the TinyMT first for test frames to get an idea for how it all works.
+
+1. Choose a test frame to be your target frame.
+
+   - You can do this by searching for any frame you want within the `Wild RNG` Tab of 3DSRNGTool.
+   - Do not worry about nature if using Sync as this will be done with TinyMT.
```
zaksabeast: This sentence is slightly confusing to me - it reads like, "Let's do TinyMT, but don't worry about nature if you're using sync, because we'll do this when we do TinyMT"

It feels like a teacher saying, "Don't worry about this math problem; we'll learn it in math class" if you're sitting in a math class.

Is this trying to say "TinyMT will always use a sync nature"?

guides/xy/en/Friend Safari RNG Guide.md
```diff
@@ -0,0 +1,104 @@
...
+   - If you accidentally trigger an encounter, afterwards walk at least 3 steps or else you won't be able to trigger another encounter when you want to.
+
+## Hitting the Target Frame
+
+1. When you are getting close to your target frame, close the menu, and then pause the game with `Start + Select`.
+
+   - Advance frame by frame by pressing `Select` until you are on your target frame.
+   - Make sure you are standing in a grass tile while doing this.
+
+2. At the same time, both unpause with `A` and turn in any direction with the D-Pad to trigger the encounter.
```
zaksabeast: Let's remove the word `both` from here since it doesn't add value to the sentence (the phrase `At the same time` implies you'll do both the 'A' and the 'D-Pad' presses):
```
At the same time, unpause with `A` and turn in any direction with the D-Pad to trigger the encounter.
```

guides/xy/en/Friend Safari RNG Guide.md
```diff
@@ -0,0 +1,104 @@
...
+
+3. Calibrate delay if needed, although delay of 6 should work for Friend Safari.
```
zaksabeast: Please add `a` in front of `delay`:
```
Calibrate delay if needed, although a delay of 6 should work for Friend Safari
```

guides/xy/en/Friend Safari RNG Guide.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Reseeding Method
+
+Once you have the hang of manipulating the TinyMT to correspond to a target frame, then you can go for higher frames. Using NTR Helper and reseeding you can search for a target frame faster.
+
+1. Load up XY while holding the `Left` D-pad. The screen should flash 3 times.
+
+2. Press `One Click` in NTR Helper to connect and grab the initial seed.
```
zaksabeast: These first two points feel like NTR Helper instructions - would it make sense to replace [line 18 of this guide](https://github.com/zaksabeast/PokemonRNGGuides/pull/94/files#diff-fbbd604c82b4d47ffc037271bbeae3c6R18) with these two steps?

guides/xy/en/Friend Safari RNG Guide.md
```diff
@@ -0,0 +1,104 @@
+# XY Friend Safari RNG Guide
+
+## Tools
+
+    - [3DSRNGTool](https://github.com/wwwwwwzx/3DSRNGTool/releases)
+
+    - [PCalc for XY](https://pokemonrng.com/downloads/pcalc/pcalc-xy)
```
zaksabeast: This link is invalid - `https://pokemonrng.com/downloads/pcalc/pcalc-xy.zip` is the correct one.

guides/xy/en/Friend Safari RNG Guide.md
ShinySylveon04: I've moved that section to be after the `Required Reading` section. 

Perhaps we don't need the `Required Reading` section? It's more of extra info on TinyMT, but I don't think it's directly needed for this guide.

guides/xy/en/Friend Safari RNG Guide.md
ShinySylveon04: The order of directions is:
- Find target frame in main window (requires setting up initial search, which is why `Sync` is mentioned here, but it isn't a part of the search criteria for this step and shouldn't be used)
- Navigate to Tiny Timeline Tool to do the actual TinyMT manipulation (here is where `Sync` will be used, but it may be confusing as to why there's a `Sync` section in the main window and not here)

Perhaps this entire section should be expanded upon to explain what to search for here and what will be manipulated with TinyMT instead.

```1. Choose a test frame to be your target frame.

   - You can do this by searching for any frame you want within the `Wild RNG` Tab of 3DSRNGTool.
   - Do not worry about nature if using Sync as this will be done with TinyMT.
   - Right click on the frame you want and click "Set as Target Frame".```

File: guides/xy/en/Friend Safari RNG Guide.md (hunk 8 above)
Comment (ShinySylveon04) [reply guides/xy/en/Friend Safari RNG Guide.md]: I would agree, except the NTR Helper instructions has more details than just those two steps that new users may need. However, I believe instead we should have a separate guide for NTR Helper and link to the relevant section (i.e. How to connect to NTR Helper).

But for now, I have replaced those instructions with:

>1. Load up XY and connect to NTR Helper within 3DSRNGTool.

 

> 

  - Press `One Click` in NTR Helper to connect and grab the initial seed.
>    - Directions for how to do this can be found in the NTR Helper Usage link above.

File: guides/xy/en/Friend Safari RNG Guide.md (hunk 5 above)
Comment (ShinySylveon04) [reply guides/xy/en/Friend Safari RNG Guide.md]: >Do not check the `Sync` box even if you are using sync as this will be done later with TinyMT.

Quoted from Zak - does this fix the above confusion?

File: guides/xy/en/Friend Safari RNG Guide.md (hunk 8 above)
Comment (ShinySylveon04) [reply guides/xy/en/Friend Safari RNG Guide.md]: In relation to this, instructions for enabling debugger in NTR and Rosalina should be added (though this will be in the connecting to NTR Helper section).

Also, mention that PSS should be disabled in the game.

File: guides/xy/en/Friend Safari RNG Guide.md [hunk 10]
```diff
@@ -0,0 +1,104 @@
...
+## Required Reading
+
+    - [TinyMT Timeline](https://github.com/wwwwwwzx/3DSRNGTool/wiki/Gen6-TinyMT-Timeline-Calibration)
+    - [NTR Helper Usage](https://github.com/wwwwwwzx/3DSRNGTool/wiki/NTR-Helper-Usage)
+
+## Explanation of TinyMT frames within Tiny Timeline Tool
+
+    - **Main RNG Frame**: The main RNG frame ranges that correspond to each TinyMT frame that is listed.
+        - Through manipulation of the TinyMT timeline, you can make the TinyMT frames overlap with different main RNG frames.
+    - **Enctr?**: Determines if there will be an encounter or not if the character turns or walks into the grass during those TinyMT seeds.
```
RichardPaulAstley: Should be highlighted as the MAIN factor of FS RNG. And Enctr? MUST match this in order to avoid any mistake / fail since this is the most important thing to calibrage.

guides/xy/en/Friend Safari RNG Guide.md
```diff
@@ -0,0 +1,104 @@
...
+   - Make sure the `Consider Delay` box is checked and delay is 6.
+
+5. Click on `Calibrate` by the `Tiny Seeds` and wait for it to calibrate for you.
+
+6. If the TinyMT frames you want do not overlap with the target frame, take a step out of the grass and then back into the grass, open menu, then calibrate like before.
+
+   - Repeat as needed until the TinyMT frames you want line up with the target frame.
+   - To advance the TinyMT frames a smaller amount you can just turn in place once instead of taking steps.
+   - To advance frames faster you can walk up and down the left side of the area since steps advance TinyMT frames the fastest.
+   - If you accidentally trigger an encounter, afterwards walk at least 3 steps or else you won't be able to trigger another encounter when you want to.
```
RichardPaulAstley: Had issues with 3 steps. 5 is more safe iirc. Maybe add a note to do that after every successful RNG ? (it's quite obvious, but we never know)

guides/xy/en/Friend Safari RNG Guide.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Hitting the Target Frame
+
+1. When you are getting close to your target frame, close the menu, and then pause the game with `Start + Select`.
```
RichardPaulAstley: 200 frames is a good range, to avoid X+2 advancement

guides/xy/en/Friend Safari RNG Guide.md
```diff
@@ -0,0 +1,104 @@
...
+
+   - Advance frame by frame by pressing `Select` until you are on your target frame.
+   - Make sure you are standing in a grass tile while doing this.
+
+2. At the same time, unpause with `A` and turn in any direction with the D-Pad to trigger the encounter.
+
+3. Calibrate a delay if needed, although delay of 6 should work for Friend Safari.
```
RichardPaulAstley: 6 frames for a turn, 14 frames for a step.

guides/xy/en/Friend Safari RNG Guide.md
RichardPaulAstley: Update : To 100% secure it, you MUST close the menu in the TinyMT Index corresponding to your target.

guides/xy/en/Friend Safari RNG Guide.md
RichardPaulAstley: Update : Must be UNDER (and not under and equal) to 5 to secure an encounter.

Issue #106, opened 2020-06-30
[Site Rework #2] Separating games and tools (emus, download links to RNG Abuse tool etc)


RichardPaulAstley:
Rather than listing everything as a whole, it'd be cool to have a first part focusing on ressources (tools, emulators, download links, setup etc) and a second part with games' guides

![image](https://user-images.githubusercontent.com/25870563/86136788-c2b65980-baec-11ea-9143-3c115799ee26.png)

Like this

Issue #107, opened 2020-06-30
[Site Rework #3] Ordering games by release date


RichardPaulAstley:
Every pokemon website uses the release date order. Also since many RNG mechanics are shared, it'd also help to know which kind of game / rng you'll check guides of.

If you regroup them by generation, they can be ordered in an alphabetically. Example :

Gen 1 - Red, Blue, Yellow
Gen 2 - Gold, Silver, Crystal
Gen 3 - Emerald
Gen 3 - Fire Red, Leaf Green
Gen 3 - Ruby, Sapphire
Gen 4 - Diamant, Pearl, Platinum
Gen 4 - Heart Gold, Soul Silver

etc

PR #129, opened 2021-03-06
Change guide categories

ShinySylveon04:
Guide categories are now by generation instead of alphabetical. Also, some categories have been renamed to reflect the official names of the games.

@zaksabeast will need to review this PR before it gets merged

This will close issues #106 and #107.


src/utils/sort-guide-list.js
```diff
@@ -0,0 +1,19 @@
+const list = [
```
zaksabeast, line 1: Let's rename this to `sortedCategories`.

src/utils/sort-guide-list.js
```diff
@@ -0,0 +1,19 @@
...
+  'X and Y',
+  'Omega Ruby and Alpha Sapphire',
+  'Sun and Moon',
+  'Ultra Sun and Ultra Moon',
+  'Sword and Shield',
+];
+
+export const sortList = (first, second) => {
+  return list.indexOf(first) - list.indexOf(second);
+};
```
zaksabeast, line 19: This error will prevent building the site if a category is added or renamed without being added to the sorting list.  During development, the error will show up in place of the site to let the developer or guide editor know the list should be updated.

```suggestion
export const sortList = (first, second) => {
  const firstIndex = list.indexOf(first);
  const secondIndex = list.indexOf(second);

  if (firstIndex === -1 || secondIndex === -1) {
    // This will stop the Gatsby release build or show an error during development
    // to make sure unordered categories won't deploy.
    throw new Error(
      'Unexpected category!  Please ensure all guide categories are in the sorted category list.',
    );
  }

  return firstIndex - secondIndex;
};
```

src/utils/sort-guide-list.js
```diff
@@ -0,0 +1,19 @@
...
+  'Black and White',
+  'Black 2 and White 2',
+  'X and Y',
+  'Omega Ruby and Alpha Sapphire',
+  'Sun and Moon',
+  'Ultra Sun and Ultra Moon',
+  'Sword and Shield',
+];
+
+export const sortList = (first, second) => {
```
zaksabeast, line 17: Let's rename this to be `sortCategories`.

zaksabeast (changes requested):

zaksabeast:
Oh one more thing, we should rename the file to `sort-categories.js` since we're sorting categories instead of guides, and it's the name of the util being exported.

zaksabeast (approved):

PR #155, opened 2021-08-30
Add notifications

ShinySylveon04:
This PR adds notifications to the site so visitors can see what the latest changes are.


RichardPaulAstley (approved):

src/components/notifications.js
```diff
@@ -0,0 +1,71 @@
...
+  ? localStorage.getItem('lastSeenNotification')
+  : null;
+const lastSeenNotification = storageItem ?? '2000-1-1';
+const messageList = orderBy(messages, ['date'], ['desc']);
+const count = filter(messageList, (message) => {
+  return message.date > lastSeenNotification;
+}).length;
+
+export const Notifications = () => {
+
```
zaksabeast, line 24: Let's remove this empty line.

src/components/notification-list.js
```diff
@@ -0,0 +1,66 @@
...
+  divider: {
+    margin: theme.spacing(1, 0),
+  },
+}));
+
+export const NotificationList = ({messageList}) => {
+    const classes = useStyles();
+    return (
+        <Paper className={classes.paper}>
+                <List className={classes.list}>
```
zaksabeast, line 31: It looks like `yarn format` needs to be run.

src/components/notifications.js
```diff
@@ -0,0 +1,71 @@
...
+import orderBy from 'lodash/orderBy';
+import filter from 'lodash/filter';
+import { NotificationList } from './notification-list';
+
+const isBrowser = typeof window !== 'undefined';
+const storageItem = isBrowser
+  ? localStorage.getItem('lastSeenNotification')
+  : null;
+const lastSeenNotification = storageItem ?? '2000-1-1';
+const messageList = orderBy(messages, ['date'], ['desc']);
```
zaksabeast, line 18: I don't think the args need to be arrays.
```suggestion
const messageList = orderBy(messages, 'date', 'desc');
```

src/components/notification-list.js
```diff
@@ -0,0 +1,66 @@
...
+                                month: 'long',
+                                day: 'numeric',
+                              },
+                            )}
+                          </Typography>
+                        )}
+                      </ListItem>
+                      {index < messageList.length - 1 ? (
+                        <Divider className={classes.divider} />
+                      ) : null}
```
zaksabeast, line 60: We don't need to return null:
```suggestion
                      {index < messageList.length - 1 && (
                        <Divider className={classes.divider} />
                      )}
```

src/components/notification-list.js
```diff
@@ -0,0 +1,66 @@
...
+                        {message.date && (
+                          <Typography variant="caption" color="textSecondary">
+                            {new Date(message.date).toLocaleDateString(
+                              'en-US',
+                              {
+                                year: 'numeric',
+                                month: 'long',
+                                day: 'numeric',
+                              },
+                            )}
```
zaksabeast, line 54: Pulling this out as a named function could help make it easier to read this component when skimming through it.

src/components/notification-list.js
```diff
@@ -0,0 +1,66 @@
...
+                  {messageList.map((message, index) => (
+                    <React.Fragment key={message.title}>
+                      <ListItem
+                        alignItems="flex-start"
+                        className={classes.listItem}
+                      >
+                        <Typography gutterBottom>{message.title}</Typography>
+                        <Typography gutterBottom variant="body2">
+                          <span
+                            id="notification-message"
```
zaksabeast, line 41: Let's remove this Id since Ids should be unique and we're going to be rendering multiple components that use it.

src/components/notifications.js
```diff
@@ -0,0 +1,71 @@
...
+        placement="bottom-start"
+        transition
+        anchorEl={anchorRef.current}
+        disablePortal
+      >
+        {({ TransitionProps }) => (
+          <ClickAwayListener
+            onClickAway={() => {
+              setPopperOpen(false);
+            }}
```
zaksabeast, line 61: Not a necessary change, but it would help make the code a bit more succinct.
```suggestion
            onClickAway={() => setPopperOpen(false)}
```

src/components/notification-list.js
```diff
@@ -0,0 +1,66 @@
...
+
+export const NotificationList = ({messageList}) => {
+    const classes = useStyles();
+    return (
+        <Paper className={classes.paper}>
+                <List className={classes.list}>
+                  {messageList.map((message, index) => (
+                    <React.Fragment key={message.title}>
+                      <ListItem
+                        alignItems="flex-start"
```
zaksabeast, line 35: Since this prop is style related, I think we should move this to `classes.listItem`.  Props like this are convenient to avoid the boilerplate of `classes`, but it's confusing to mix both props and classes.

For example, if someone adds `alignItems: 'flexEnd'` to `classes.listItem`, which style will be used?  One of the two places wouldn't work as expected, and it might not be obvious where to look for this type of bug.

zaksabeast (changes requested):

src/components/notification-list.js
```diff
@@ -0,0 +1,61 @@
...
+
+export const NotificationList = ({ messageList }) => {
+  const classes = useStyles();
+  const notificationDate = date => {
+    return new Date(date).toLocaleDateString('en-US', {
+      year: 'numeric',
+      month: 'long',
+      day: 'numeric',
+    });
+  };
```
zaksabeast, line 36: Oh sorry, I meant pull this outside of the component.

When React checks which components it needs to update, it will check if a component's props have updated by comparing the current props to the previous props.  When a function lives inside of a component, a new function is effectively created each time the component is rendered, which means the function has a new reference in memory.  If this function is used as a prop for a component, it will always appear as if the props have changed, which will hurt performance.

As a result, it's generally good practice for functions to live outside of components when possible to prevent this from happening.  That won't happen in this case, but it's good to exercise that muscle when possible so it's instinctive when that situation does come up.

Also, there's not really a need for it to exist inside of the component, so the component itself will have less code.

src/components/notification-list.js
```diff
@@ -0,0 +1,61 @@
...
+              <Typography gutterBottom>{message.title}</Typography>
+              <Typography gutterBottom variant="body2">
+                <span dangerouslySetInnerHTML={{ __html: message.text }} />
+              </Typography>
+              {message.date && (
+                <Typography variant="caption" color="textSecondary">
+                  {notificationDate(message.date)}
+                </Typography>
+              )}
+            </ListItem>
```
zaksabeast, line 52: Not necessary, but pulling this into a `NotificationListItem` or `NotificationItem` component might be good for a separation of concerns.

zaksabeast (changes requested):
Looks good, getting super close!

zaksabeast (approved):

PR #168, opened 2021-09-20
Update the PCalc install guide for Luma

ShinySylveon04:
Since the 11.16 3DS update, the official Luma no longer supports NTR. A way around this is to use a Luma fork that is compatible with NTR. The guide has been updated to reflect this change.


zaksabeast (approved):

RichardPaulAstley:
(I'll let Zak review that since it's more typos check)

PR #171, opened 2021-09-27
Use MUI divider to separate guides

ShinySylveon04:
Currently the CSS is using a border to divide guides on the category pages. This PR changes the divider to instead use material-ui's divider component.

Current view
image

Changed view
image

Take note of the last guide no longer having a divider after it.

### DIFF
--- a/src/components/guide-list.js
+++ b/src/components/guide-list.js
@@ -7,6 +7,7 @@ import ListItem from '@material-ui/core/ListItem';
 import ListItemText from '@material-ui/core/ListItemText';
 import VisibilityIcon from '@material-ui/icons/Visibility';
 import IconButton from '@material-ui/core/IconButton';
+import Divider from '@material-ui/core/Divider';
 import orderBy from 'lodash/orderBy';
 
 const useStyles = makeStyles(theme => ({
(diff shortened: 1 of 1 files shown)


src/components/guide-list.js
```diff
@@ -32,31 +35,36 @@ export const GuideList = ({ subCategoryTitle, guides }) => {
...
-          <Typography variant="h5" component="p" id={guideTitleId}>
-            {title}
-          </Typography>
-          <Typography variant="body2" id={guideDescriptionId}>
-            {description}
-          </Typography>
-        </ListItemText>
-      </ListItem>
+      <React.Fragment key={index}>
+        <ListItem className={classes.guideInfo} key={title} disableGutters>
```
zaksabeast, line 45: Let's move the key:
```suggestion
      <React.Fragment key={title}>
        <ListItem className={classes.guideInfo} disableGutters>
```

zaksabeast (changes requested):

src/components/guide-list.js
```diff
@@ -32,31 +32,36 @@ export const GuideList = ({ subCategoryTitle, guides }) => {
...
+            <Typography variant="h5" component="p" id={guideTitleId}>
+              {title}
+            </Typography>
+            <Typography variant="body2" id={guideDescriptionId}>
+              {description}
+            </Typography>
+          </ListItemText>
+        </ListItem>
+        {index < orderedGuides.length - 1 && (
+          <Divider/>
```
zaksabeast, line 62: A bottom border might have previously been used to provide a visual affect without impacting screen readers.

Let's add `aria-hidden="true"` to be safe:
```suggestion
          <Divider aria-hidden="true" />
```

Aria rules are used a bit different than regular react props.  [Related docs can be found here](https://reactjs.org/docs/accessibility.html).

zaksabeast (changes requested):

PR #186, opened 2021-12-06
Add CaptureSight install guide

ShinySylveon04:
This PR adds a guide to installing CaptureSight for use with SwSh and BDSP RNG.

### DIFF
--- a/guides/Tools and Emulators/CaptureSight Install.md
+++ b/guides/Tools and Emulators/CaptureSight Install.md
@@ -1,14 +1,13 @@
 ---
 title: 'How to Install CaptureSight'
 description: 'Installing a tool on your Switch to help RNG Pokemon'
-slug: 'misc-switch-install-capturesight'
+slug: 'install-capturesight'
 subCategory: 'Switch'
 ---
 
 ## Tools
 
-- A Switch with Atmosphere CFW (Custom Firmware)
-  - https://switchgui.de/switch-guide/ has instructions for installing Atmosphere
+- A Switch with Atmosphere CFW (Custom Firmware) ([CFW Install Guide](https://switchgui.de/switch-guide/))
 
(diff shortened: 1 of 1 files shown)


guides/Tools and Emulators/CaptureSight Install.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Using CaptureSight
+
+1. To open CaptureSight, first start Sword/Shield or Brilliant Diamond/Shining Pearl.
+2. Press `L + D-pad Down + Click in right joystick` to toggle the Tesla Menu overlay.
+3. Open CaptureSight overlay.
+
+## Troubleshooting
+
+### Error: `dmnt:cht is not running`
```
RichardPaulAstley, line 36: I'd not put so big headers (when checking the file)

RichardPaulAstley (approved):

guides/Tools and Emulators/CaptureSight Install.md
```diff
@@ -0,0 +1,52 @@
...
+
+- Make sure to launch the game first before trying to run CaptureSight. Check that you are not also using sysbot-base. These two cannot be used at the same time as they conflict with each other.
+
+### Error: `Unsupported game update!`
+
+- Updating the game to the latest version should fix this error.
+
+### Error: `Unsupported game! Title Id: <titld_id>`
+
+- You are not using a game that CaptureSight supports. The list of supported games are as follows:
```
zaksabeast, line 50: I'm not sure if 'as follows' is necessary:
```suggestion
- You are not using a game that CaptureSight supports. The list of supported games are:
```

Should we send users to the CaptureSight readme to see which games area available, so there's only one place to update?  I'm good if we want the list to live here, but updating one place is less work than two.

zaksabeast (changes requested):

PR #188, opened 2021-12-11
Added some more info based on discoveries

albertodev01:
# Guide Details

## I want to (select one)

- [ ] Add a guide
- [x] Modify an existing guide

## Games affected

- BSDP

## Type of RNG

Stationary legends on BSDP

## Language

English

## Notes

My game is in italian so I didn't put the dialog box text


zaksabeast:
I squashed 42f73f6 and d5d0e4a along with a few formatting changes into @ShinySylveon04's PR with f3b51de.

@RichardPaulAstley had concerns with 8ec95f3 since the varied delays might be caused by character fidgets and blinks rather than being the base delay.  As a result, I'm going to leave it out for the moment.

Since these changes have been pulled into @ShinySylveon04's PR, which is going to be used as the source of truth, I'm going to close this PR.

Thank you for helping!

PR #223, opened 2023-05-10
Update Transporter Guide

ShinySylveon04:
This updates the guide for the following:
- Using the offline patch for a stable delay
- Replace PCalc with PokeReader
- Add Chatot as another tool option


zaksabeast:
Added in 0abbfb95780b13552201a908ea1963c3f6c5b656.

PR #236, opened 2025-03-27
Chinese translation

HaKu76:
Chinese translation


HaKu76 (commented):
Chinese translation

HaKu76 (commented):
Chinese translation

HaKu76 (commented):
PokeReader 3DS Chinese translation

src/components/languageButton.tsx
```diff
@@ -50,6 +57,10 @@ export const LanguageButton = ({ enSlug, esSlug }: Props) => {
         { key: "es", esSlug: P.not(P.nullish) },
         (matched) => matched.esSlug,
       )
+      .with(
+        { key: "zh", zhSlug: P.not(P.nullish) },
+        (matched) => matched.zhSlug,
+      )
       .otherwise(() => enSlug);
     setLocation(slug);
   };
```
zaksabeast, line 66: The language button currently shows all three available languages, even if only two are available.

Let's add this to the component to make sure only available languages can be chosen:
```tsx
  const availableLanguages = languages.filter((lang) => {
    return match(lang.key)
      .with("en", () => true)
      .with("es", () => esSlug != null)
      .with("zh", () => zhSlug != null)
      .otherwise(() => false);
  });

  return (
    <Flex>
      <Dropdown
        menu={{
          items: availableLanguages,
```

zaksabeast (changes requested):
Thank you!  Just a few small changes and it should be good to merge.

---

One more change that wasn't mentioned in the other comments: `build-guides.ts` needs to be updated.

Can you replace the `titleAndDescriptionChars` variable in `build-guides.ts` with this:
```
// Only letters, numbers, spaces, the en-dash, period, hyphen, é, &, /, (, ), !, %, ,, and Chinese characters
const titleAndDescriptionChars = /^[A-Za-z0-9 –.\-—é&/()!%,\u4e00-\u9fff]+$/;
```

Then try running `bun run build:guides`.  This will also automatically generate the `__generated__/guides.ts` file and format markdown files to have specific styles.

HaKu76 (commented):
I fixed some irregularities as requested

src/components/languageButton.tsx
```diff
@@ -25,40 +25,41 @@ const languages: LanguageItem[] = [
...
-        { key: "en", enSlug: P.not(P.nullish) },
-        (matched) => matched.enSlug,
-      )
-      .with(
-        { key: "es", esSlug: P.not(P.nullish) },
-        (matched) => matched.esSlug,
-      )
-      .otherwise(() => enSlug);
-    setLocation(slug);
-  };
```
zaksabeast, line 55: Whoops - this was accidentally removed.  We need to make sure both `setLanguageKey` and `availableLanguages` are in the file.

zaksabeast (changes requested):
Three changes are needed:
- [Allow Chinese characters in `build-guides.ts`](https://github.com/zaksabeast/PokemonRNGGuides/pull/236#pullrequestreview-2727405758)
- [Replace Spanish credits with English](https://github.com/zaksabeast/PokemonRNGGuides/pull/236#discussion_r2019830334)
- The new comment to restore `setLanguageKey`

zaksabeast:
Thank you for the translations!  I'm going to merge these into `chinese-translations` and make a commit to address the PR comments.  Then I'll merge `chinese-translations` into the site.

PR #243, opened 2025-04-05
Add Chinese translation for dppt wild

HaKu76:
Add Chinese translation for dppt wild


zaksabeast (changes requested):
This is great, thank you!  And it looks like the Trophy Garden portion is slightly updated - that's helpful.

One tiny change: could you run `bun run format` to format the file?  After that, this can be merged!

zaksabeast (approved):
Thank you!

PR #245, opened 2025-04-06
Create re-usable fields for PkmFilter 

SteveCookTU:


src/rngToolsUi/gen3/staticGenerator.tsx
```diff
@@ -360,9 +298,7 @@ export const Static3Generator = ({ game = "emerald" }: Props) => {
           gender:
             opts.filter_gender === "None" ? undefined : opts.filter_gender,
           ability:
-            opts.filter_ability === "None"
-              ? undefined
-              : (fromDecimalString(opts.filter_ability) ?? undefined),
+            opts.filter_ability === "None" ? undefined : opts.filter_ability,
```
zaksabeast, line 301: Not required in this PR, but wanted to share this thought to let it simmer:

At some point we should probably create a util to handle this:
```ts
import * as tst from "ts-toolbelt";

export type Optional<T> = tst.U.Replace<T, undefined, "None">;

export const fromOptionalInput = <T>(value: T | "None"): T | undefined => {
  return value == "None" ? undefined : value;
};
```

Or even better, we might define the form state as a zod schema so we can get define the `FormState` type and runtime validator at the same time.  Then we just need to `FormState.parse` instead of handling every field individually.

src/components/select.tsx
```diff
@@ -58,7 +58,7 @@ type FormikSelectValue<
   FormState extends GenericForm,
   FieldKey extends keyof FormState,
 > = FormState[FieldKey] extends string
-  ? { label: string; value: FormState[FieldKey] }[]
+  ? { label: string; value: FormState[FieldKey] | null }[]
```
zaksabeast, line 61: How come this was updated to be nullable?  I'm not seeing any type errors after removing this.

Wouldn't setting the form state prop being nullable satisfy this where needed?

zaksabeast (commented):
One quick question, otherwise this looks great!

src/components/select.tsx
SteveCookTU: Good catch, this was accidentally leftover from my previous version. Removed!

zaksabeast (approved):

PR #247, opened 2025-04-07
Chinese translation of four Emerald Guides

HaKu76:
- RNG Info
- Reseed using paintings
- Egg RNG
- Wild Retail RNG


zaksabeast (approved):
Thank you!

PR #276, opened 2025-05-03
Fix for too many args warning

sorendog:
Implements a Struct called DivParams to lower parameters in celebi rng function and updates typescript to accept the change as well. no change to functions, just moving variables around basically.

I would like to do this to gen 2 starters as well before merging as i imagine the generators have some overlap and i didn't test to see if that broke the gen 2 starters rng page but since the celebi page works now I would appreciate feedback on if this works and looks alright. To make sure it didn't cause any change to how functions were being done i did test that in my local environment it produces the same advance and state variables as what the site does currently.

I did make sure to run cargo test and etc as well but if it looks like i missed anything please let me know.


sorendog:
oh gosh all my commits are on this. i will figure out what i did and fix it but it might take a few moments...

PR #277, opened 2025-05-03
Fix for too many parameters warning

sorendog:
Implements a Struct called DivParams to lower parameters in celebi rng function and updates typescript to accept the change as well. no change to functions, just moving variables around basically.

I can quickly add this to gen 2 starters as well before merge if wanted which would remove another warning from cargo clippy.

To make sure it didn't cause any change to how functions were being done i did test it in my local environment, that it produces the same advance and state information as what the site does currently.

I did make sure to run cargo test and etc as well but if it looks like i missed anything please let me know.

Also apologies for the double pull request for same thing. Closed out the other so only the related commits would be on this one.

### DIFF
--- a/src/rngToolsUi/gen2/crystalPokemon.tsx
+++ b/src/rngToolsUi/gen2/crystalPokemon.tsx
@@ -8,7 +8,7 @@ import {
   RngToolSubmit,
   Field,
 } from "~/components";
-import { rngTools, type Gen2Spread } from "~/rngTools";
+import { rngTools, type Gen2Spread, type DivParams } from "~/rngTools";
 import { useTranslator, Translations, Translator } from "~/utils/siteLanguage";
 import { z } from "zod";
 import { HexSchema } from "~/utils/number";
@@ -179,14 +179,6 @@ type Props = {
   language: keyof typeof translations;
 };
 
-export interface DivParams {
-  adiv: number;
-  sdiv: number;
-  adiv_index: number;
-  sdiv_index: number;
-  state: number;
-}
-
 export const Gen2PokemonRng = ({ type, language }: Props) => {
   const t = useTranslator(translations, language);
   const fields = React.useMemo(() => getFields(t), [t]);
@@ -199,7 +191,7 @@ export const Gen2PokemonRng = ({ type, language }: Props) => {
           ? rngTools.crystal_generate_starters
           : rngTools.crystal_generate_celebi;
 
-      const config = {
+      const config: DivParams = {
         adiv: opts.div >>> 8,
         sdiv: opts.div & 0xff,
         adiv_index: opts.adivIndex,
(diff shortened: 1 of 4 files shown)


src/rngToolsUi/gen2/crystalPokemon.tsx
```diff
@@ -179,25 +179,35 @@ type Props = {
   language: keyof typeof translations;
 };
 
+export interface DivParams {
+  adiv: number;
+  sdiv: number;
+  adiv_index: number;
+  sdiv_index: number;
+  state: number;
+}
```
zaksabeast, line 188: We can actually import this from `~/rngTools`!

When we run `bun run dev` it builds the rust lib and generates TypeScript types.

So we can do this:
```ts
import { rngTools, type Gen2Spread, type DivParams } from "~/rngTools";
```

---

In this case `DivParams` is unused, so we have two options:
1. Type the new `config` type below with a `DivParams` imported from `~/rngTools`
2. Remove this and not import `DivParams` since `generator` will give us type errors if we pass an incorrect value

Both are perfectly valid.

src/rngToolsUi/gen2/crystalPokemon.tsx
```diff
@@ -179,25 +179,35 @@ type Props = {
...
+        sdiv_index: opts.sdivIndex,
+        state: opts.state,
+      };
       const results = await generator(
-        opts.div >>> 8,
-        opts.div & 0xff,
-        opts.adivIndex,
-        opts.sdivIndex,
-        opts.state,
+        config,
```
zaksabeast, line 210: When I run `bun run lint`, I get a type error on this line because the Celebi generator was updated, but not the Starter generator.

> I can quickly add this to gen 2 starters as well before merge if wanted which would remove another warning from cargo clippy.

Go for it!  That would also solve the type issue.

zaksabeast (changes requested):

src/rngToolsUi/gen2/crystalPokemon.tsx
sorendog: I would like to do importing of DivParams but when I do and make the adjustments needed in the code bun run lint gives me the following error. 

 `error TS2345: Argument of type 'DivParams' is not assignable to parameter of type 'number & { adiv: number; sdiv: number; adiv_index: number; sdiv_index: number; state: number; }'.`

Do you have any suggestions maybe of what I can do to fix it?

src/rngToolsUi/gen2/crystalPokemon.tsx
sorendog: Awesome! most recent commit implemented this so gen 2 starters is all done as well.

src/rngToolsUi/gen2/crystalPokemon.tsx
zaksabeast: Looks like you got it working 👍

zaksabeast (approved):
Thank you!  It's great to see that clippy number go down 😄

PR #281, opened 2025-05-04
TS for transporter generator

sorendog:
add files for typescript connections to add transporter

### DIFF
--- a/guides/Transporter/Transporter.mdx
+++ b/guides/Transporter/Transporter.mdx
@@ -33,7 +33,7 @@ RNGing with Transporter has a delay between pressing `A` and Pokémon generation
    - Example: If you started on frame 1000 and hit frame 1197, your delay is 197.
 1. When asked to transfer the Pokémon, click "No."
 1. Repeat the process 5-10 times to get 5-10 delays.
-1. Input the most frequent delay or an average into 3DSRNGTool's "Delay" box.
+1. Input the most frequent delay or an average into the "Delay" box.
 
 ## Step 2: Find a spread to RNG
 
@@ -46,16 +46,28 @@ Follow these steps in with the tool below to find a desired spread:
 1. Change the "Max Advances" box to the maximum frame you want to search.
 1. Click "Generate."
 
-All results are potential wanted frames. Pick your favorite!
+All results are potential wanted advances. Pick your favorite!
 
 ## Step 3: Hit the target
 
 1. Wait at the game selection screen until you get close to your wanted frame.
 1. Once close, press `Start + Select` to pause the game.
-1. Press `Select` to advance the frames one by one until you reach your wanted frame.
+1. Press `Select` to move the advances one by one until you reach your wanted frame.
 1. Once at your wanted frame, press and hold `A`.
 1. Check if you got the Pokémon you wanted; if not, don’t transfer your Pokémon and try again.
 
 This may take several attempts.
+
+Transporter generates 20 Pokémon at once. If you have 20 of the same Pokémon or cloned, you can have 20 chances to RNG the same Pokémon.
+
+1. Catch 20 of the same Pokémon (or clone them).
+1. Make sure all 20 are in Box 1 of your Gen 1/2 game.
+
+In the tool below:
+
+1. Change the "Filters" box to have the IVs you want to RNG.
+1. Search for spreads using the initial seed PokeReader shows.
+1. Ensure the number of consecutive advances is at least 10.
+1. Aim for the middle advance as your target advance.
 
 <Transporter />
--- a/guides/Transporter/Transporter.mdx
+++ b/guides/Transporter/Transporter.mdx
@@ -10,7 +10,6 @@ tag: "cfw"
 ## Tools
 
 - [A 3DS with PokeReader](/install-pokereader)
-- [3DSRNGTool](https://github.com/wwwwwwzx/3DSRNGTool/releases)
 
 ## Useful note
 
@@ -24,15 +23,12 @@ RNGing with Transporter has a delay between pressing `A` and Pokémon generation
 
 1. Make sure you have at least one Pokémon in Box 1, Slot 1 of your Gen 1/2 game.
 1. Load Pokémon Transporter with PokeReader on your 3DS.
-1. Open 3DSRNGTool on your computer.
-1. Change 3DSRNGTool's "Game Version" to "Transporter."
-1. Change 3DSRNGTool's "Seed" to the Initial Seed PokeReader shows.
-1. Change 3DSRNGTool's tab to "Stationary RNG."
+1. Enter your Initial seed in the tool below.
 1. In Transporter, press `Start` to select the game for Transporter.
 1. Press `Start + Select` to pause the game.
 1. Note your current frame; this is your starting frame.
 1. Press `A` to unpause and let Transporter generate the Pokémon.
-1. Search for the exact IVs PokeReader indicates using 3DSRNGTool to find your hit frame.
+1. Search for the exact IVs PokeReader indicates using the tool below to find your hit frame.
 1. Subtract your hit frame from your starting frame to get your delay.
    - Example: If you started on frame 1000 and hit frame 1197, your delay is 197.
 1. When asked to transfer the Pokémon, click "No."
@@ -41,15 +37,14 @@ RNGing with Transporter has a delay between pressing `A` and Pokémon generation
 
 ## Step 2: Find a spread to RNG
 
-Follow these steps in 3DSRNGTool to find a desired spread:
+Follow these steps in with the tool below to find a desired spread:
 
-1. Change the "Gender Ratio" to match the Pokémon you want to RNG.
-1. Change the "Index of Target Pkm" to "1."
-1. Change only the gender and IVs to find a spread you want.
+1. Change the "Transporter Gender" to match the Pokémon you want to RNG.
+1. Change IVs to find a spread you want.
    - Nature, ability, and shininess are determined in the Gen 1/2 game before Transporter.
-1. Change the first "Frame Range" box to your current frame.
-1. Change the second "Frame Range" box to the maximum frame you want to search.
-1. Click "Search."
+1. Change the "Initial Advance" box to your current frame.
+1. Change the "Max Advances" box to the maximum frame you want to search.
+1. Click "Generate."
 
 All results are potential wanted frames. Pick your favorite!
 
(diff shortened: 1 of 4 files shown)


src/rngToolsUi/gen6/transporter/transporter.tsx
sorendog, line 1: on line 120-ish is where the error i get comes from when i run bun run lint

sorendog:
i just realized i changed it to try to make it just work, going to do a commit in a second to change back to what i had.

src/rngToolsUi/gen6/transporter/transporter.tsx
zaksabeast: You're mostly there!

**Here's the issue:**

`transporter_genders` is an array of genders, but `FormikSelect` is a single dropdown that chooses a single gender.  The type mismatch is because `TransporterGender` doesn't equal `TransporterGender[]`.

---

**Why do we need a list of genders?**

3DSRNGTool let's people select multiple genders, because different genders take a different number of advances to generate:
Screenshot 2025-05-04 at 7 35 13 PM

So if the user has a Pokemon in slot 3, we need to know the genders leading up to that Pokemon.

---

**What's the solution?**

For this web tool, we have two options:
- Use a `FormikInput` and allow people to type a list of characters like what 3DSRNGTool has
- Assume people will only transfer one Pokemon (the one they want to RNG) and only accept a single gender

I lean towards the second option.  We recommend people duplicate their Pokemon into a few slots before transferring it in order to increase the odds one slot will hit the target, so assuming there's only a single Pokemon is safe.  It's also the better user experience.

We can do this by:
- Converting `transporter_genders` in the validator to not be an array
- Updating the initial values to not use an array
- Updating `transporter_genders: opts.transporter_genders` to `transporter_genders: [opts.transporter_genders]` since rust still expects an array
- Removing the `Target` field, `target` in the validator, and hardcode `0` as the target in the `generate_transporter` opts

---

**One more note:**

We have a `toOptions` utils that allows us to convert a list into options for `FormikSelect`.

```tsx
// Pull this out to be reused and prevent typos
const transporterGenders = [
  "NoGender",
  "RandomGender",
  "Mythical",
] as const;

// Use this in FormikSelect
const transporterGenderOptions = toOptions(transporterGenders);

// Use this in the validator (you're aleady doing this)
const TransporterGenderEnum = z.enum(transporterGenders);
```

sorendog: Thank you! This was exactly the information I needed to get this finished out, i hard coded to zero and so have it set up with the assumption people would be doing 1 at a time. I really appreciate your thorough response.

src/rngToolsUi/gen6/transporter/transporter.tsx
```diff
@@ -0,0 +1,152 @@
...
+  {
+    title: "Shiny",
+    dataIndex: "shiny",
+    render: (shiny) => (shiny ? "Yes" : "No"),
+  },
+  {
+    title: "IVs",
+    dataIndex: "ivs",
+    monospace: true,
+    render: (ivs) =>
```
zaksabeast, line 42: IV results are a bit of a special case.  When we had a poll in the Discord, people preferred them to be in separate columns.

We have some easy utils to do this:
- `flattenIvs` function
- `FlattenIvs` type
- `ivColumns` pre-built, shared columns

---

**What do the utils do?**

The `flattenIvs` takes an object and merges the ivs to the top level:
```ts
const test = flattenIvs({
  species: "Torchic",
  ivs: { hp: 31, atk: 31, def: 31, spa: 31, spd: 31, spe: 31 },
});
console.log(test);
// { species: 'Torchic', ivs: { hp: 31, atk: 31, def: 31, spa: 31, spd: 31, spe: 31 } }
```

`FlattenIvs` is the type equivalent:
```ts
type PretendState = {
  species: Species;
  ivs: Ivs;
}

type Test = FlattenIvs<PretendState>;
// Test's type is now this:
// type Test = {
//   species: Species;
//   hp: number;
//   atk: number;
//   def: number;
//   spa: number;
//   spd: number;
//   spe: number;
// }
```

---

**How can we use these tools?**


At the top of the file we can import the utils:
```tsx
import {
  flattenIvs,
  FlattenIvs,
  ivColumns,
} from "~/rngToolsUi/shared/ivColumns";
```

Then we can create a new result type:
```tsx
type Result = FlattenIvs<Stationary6State>;
```

`Result` should be used everywhere `Stationary6State` is currently used.

Next we transform the result before setting it:
```tsx
// Instead of this
// setResults(results);

// Do this
setResults(results.map(flattenIvs));
```

And lastly we replace the IV columns from this:
```tsx
  {
    title: "Shiny",
    dataIndex: "shiny",
    render: (shiny) => (shiny ? "Yes" : "No"),
  },
  {
    title: "IVs",
    dataIndex: "ivs",
    monospace: true,
    render: (ivs) =>
      [ivs.hp, ivs.atk, ivs.def, ivs.spa, ivs.spd, ivs.spe].join(" / "),
  },
```

To this:
```tsx
{
    title: "Shiny",
    dataIndex: "shiny",
    render: (shiny) => (shiny ? "Yes" : "No"),
  },
  ...ivColumns,
```

---

**What's the final result?**

| Current PR | After changes |
| - | - |
| Screenshot 2025-05-05 at 10 17 01 AM | Screenshot 2025-05-05 at 10 16 28 AM |

src/rngToolsUi/gen6/transporter/transporter.tsx
```diff
@@ -0,0 +1,152 @@
...
+  {
+    label: "Transporter Gender",
+    input: (
+      <FormikSelect<FormState, "transporter_genders">
+        name="transporter_genders"
+        options={transporterGenderOptions}
+      />
+    ),
+  },
+];
```
zaksabeast, line 115: Let's add IV filter fields in case users wants specific IVs.

In rust we have a shared `PkmFilter` struct that several of our generators use, including this one.

In TS we have shared fields that let us take advantage of this with minimal code: `getPkmFilterFields`

It's a function so we can opt out of fields.  In this case we only want to filter IVs, because shininess, gender, ability, and nature are predetermined.

```tsx
const fields: Field[] = [
  // other fields
  ...getPkmFilterFields({
    ivs: true,
    ability: false,
    gender: false,
    nature: false,
    shiny: false,
  }),
];
```

A few more changes are needed:
- Validator
- initial values
- Passing the filter to `generate_transporter`

But I'll leave those as an exercise for you to explore how we do this elsewhere.  If you get stuck, let me know!

guides/Transporter/Transporter.mdx
```diff
@@ -84,3 +84,5 @@ In 3DSRNGTool:
 1. In the "RNG Info" box of 3DSRNGTool, change "delay" to this new delay.
 1. Press "Calculate" to find your potential frames with this new delay for your target index.
 1. Find the highest frame among the 10+ consecutive frames from earlier and use it as your target frame.
+
+<Transporter />
```
zaksabeast, line 88: We should 100% add this to the guide, although as-is the guide references 3DSRNGTool.  We need to update this for the embedded tool.

If you want to update the guide, go for it!  If not, Shiny or I can do it.  Let us know which you'd prefer.

zaksabeast (changes requested):
Super close!  Just a few small tweaks.

src/rngToolsUi/gen6/transporter/transporter.tsx
zaksabeast: Sure thing!  I'm happy to help.

guides/Transporter/Transporter.mdx
sorendog: I updated the guide to reference the tool. For now I removed the section with regards to doing 20 pokemon at once. but maybe that section would do good as it's own page now that still links to 3dsrng tool since doing 20 at once right now would be best set up for in 3dsrng tool?

sorendog:
I think I got everything, so going to request re-review. If I missed anything let me know!

src/rngToolsUi/gen6/transporter/transporter.tsx
```diff
@@ -0,0 +1,169 @@
...
+      max_advances: opts.max_advances,
+      delay: opts.delay,
+      target: 0,
+      transporter_genders: [opts.transporter_genders],
+      tsv: opts.tsv,
+      filter: {
+        gender: null,
+        ability: null,
+        nature: null,
+        ivs: null,
```
zaksabeast, line 150: We need to hook up the iv filters - they don't work at the moment:
```suggestion
        ivs: {
          min_ivs: opts.filter_min_ivs,
          max_ivs: opts.filter_max_ivs,
        },
```

guides/Transporter/Transporter.mdx
ShinySylveon04, line 1: I think it would be good to change “frames” to “advances.” We’re trying to phase out the term “frames” and use “advances” instead. Both the tool and PokeReader use “advances,” so keeping the wording the same will help avoid confusion.

guides/Transporter/Transporter.mdx
```diff
@@ -41,15 +37,14 @@ RNGing with Transporter has a delay between pressing `A` and Pokémon generation
 
 ## Step 2: Find a spread to RNG
 
-Follow these steps in 3DSRNGTool to find a desired spread:
+Follow these steps in with the tool below to find a desired spread:
```
ShinySylveon04, line 40: A small typo here.
```suggestion
Follow these steps with the tool below to find a desired spread:
```

guides/Transporter/Transporter.mdx
```diff
@@ -24,15 +23,12 @@ RNGing with Transporter has a delay between pressing `A` and Pokémon generation
...
+1. Enter your Initial seed in the tool below.
 1. In Transporter, press `Start` to select the game for Transporter.
 1. Press `Start + Select` to pause the game.
 1. Note your current frame; this is your starting frame.
 1. Press `A` to unpause and let Transporter generate the Pokémon.
-1. Search for the exact IVs PokeReader indicates using 3DSRNGTool to find your hit frame.
+1. Search for the exact IVs PokeReader indicates using the tool below to find your hit frame.
 1. Subtract your hit frame from your starting frame to get your delay.
    - Example: If you started on frame 1000 and hit frame 1197, your delay is 197.
 1. When asked to transfer the Pokémon, click "No."
```
ShinySylveon04, line 34: There's a 3DSRNGTool reference two lines below this (but I can't comment on that line for some reason).

zaksabeast (changes requested):
One tiny tweak, but otherwise this is ready from a code perspective.  I'll let @ShinySylveon04 take a look at the guide.

guides/Transporter/Transporter.mdx
```diff
@@ -63,24 +58,4 @@ All results are potential wanted frames. Pick your favorite!
 
 This may take several attempts.
 
-## Alternative method with increased chances
-
-Transporter generates 20 Pokémon at once. If you have 20 of the same Pokémon or cloned, you can have 20 chances to RNG the same Pokémon.
```
ShinySylveon04, line 68: I think that's still a good part to keep, but we can update it to be a bit simpler and remove 3DSRNGTool references:
```
Transporter generates 20 Pokémon at once. If you have 20 of the same Pokémon or cloned, you can have 20 chances to RNG the same Pokémon.

1. Catch 20 of the same Pokémon (or clone them).
1. Make sure all 20 are in Box 1 of your Gen 1/2 game.

In the tool below:

1. Change the "Filters" box to have the IVs you want to RNG.
1. Search for spreads using the initial seed PokeReader shows.
1. Ensure the number of consecutive advances is at least 10.
1. Aim for the middle advance as your target advance.
```

ShinySylveon04 (changes requested):

sorendog:
All notes implemented!

zaksabeast (approved):
Thank you!  This is great!

Looks good from a code perspective - I'll wait for @ShinySylveon04's guide approval before merging.

ShinySylveon04 (approved):

PR #287, opened 2025-05-06
Add RS Retail TID Guide

ShinySylveon04:
Please review for the RS TID Tool part I have so far. ~~There's an issue with the page refreshing when navigating between tabs that I need help fixing.~~ Issue fixed, thank you!

Don't forget to add to RS SID methods.


src/rngToolsUi/gen3/rstid/rstid.tsx
```diff
@@ -0,0 +1,117 @@
...
+    input: <FormikIdFilter<FormState> name="filter" optional />,
+  },
+];
+
+export const RsTidSidGenerator = ({ handleTarget }) => {
+  const [results, setResults] = React.useState<Result[]>([]);
+
+  const onSelectTarget = (advance: number, results) => {
+    handleTarget(advance, results);
+  };
```
zaksabeast, line 74: This is a passthrough function, so we can remove this and just pass `onSelectTarget` as a prop.

src/rngToolsUi/gen3/rstid/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+  return (
+    <React.Fragment>
+      <Stepper onChange={onChange} current={current} items={items} />
+
+      {current < items.length - 1 && (
+        <React.Fragment>
+          {targetSet && <NextButton />}
+          <RsTidSidGenerator handleTarget={handleTarget} />
+          {targetSet && <NextButton />}
+        </React.Fragment>
```
zaksabeast, line 65: We usually don't want to use Fragments.  Use `Flex` here instead of a `React.Fragment`.
```tsx
        <Flex vertical>
          {targetSet && <NextButton />}
          <RsTidSidGenerator onSelectTarget={handleTarget} />
          {targetSet && <NextButton />}
        </Flex>
```

Most styling issues people run into are layout related.  Those happen when the layout of multiple components is controlled by some external source.

`React.Fragment` is the same as returning an array of components, which gives layout control to the parent.

Usually if we want to group components together, we should strictly define _how_ those components should exist together to avoid layout issues.

src/rngToolsUi/gen3/rstid/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+        Next
+      </Button>
+    </Flex>
+  );
+
+  return (
+    <React.Fragment>
+      <Stepper onChange={onChange} current={current} items={items} />
+
+      {current < items.length - 1 && (
```
zaksabeast, line 60: The reason your state disappears when you change tabs is because this 'un-renders' the `RsTidGenerator` component, which owns the state.

Instead of un-rendering it, you can hide it:
```tsx
<Flex vertical display={current < items.length - 1 ? "flex" : "none"}>
  {targetSet && <NextButton />}
  <RsTidSidGenerator onSelectTarget={handleTarget} />
  {targetSet && <NextButton />}
</Flex>
```

The state persists since it stays rendered.

src/rngToolsUi/gen3/rstid/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+    return advance;
+  };
+
+  const NextButton = () => (
+    <Flex justify="end">
+      <Button
+        trackerId="test"
+        type="primary"
+        size="middle"
+        onClick={() => next()}
```
zaksabeast, line 49: This is a passthrough we can remove:
```suggestion
        onClick={next}
```

src/rngToolsUi/gen3/rstid/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+      >
+        Next
+      </Button>
+    </Flex>
+  );
```
zaksabeast, line 54: This creates a new function every time `RsTid` is rendered, which will affect performance and cause unnecessary re-renders.

Let's pull this into an external component and pass `next` as a prop.

src/rngToolsUi/gen3/rstid/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  return (
+    <React.Fragment>
```
zaksabeast, line 57: Similar to the other comment, let's use `Flex` here.

src/rngToolsUi/gen3/rstid/rstid.tsx
```diff
@@ -0,0 +1,117 @@
...
+  Button,
+  Icon,
+} from "~/components";
+import React from "react";
+import { denormalizeIdFilter, IdFilterSchema } from "~/types/id";
+import { z } from "zod";
+
+type Result = Gen3TidSidResult & { time: string };
+
+const getColumns = (): ResultColumn<Result>[] => [
```
zaksabeast, line 18: This doesn't need to be a function
```suggestion
const columns: ResultColumn<Result>[] = [
```

src/rngToolsUi/gen3/rstid/rstid.tsx
```diff
@@ -0,0 +1,117 @@
...
+      fields={fields}
+      columns={getColumns()}
+      results={results}
+      initialValues={initialValues}
+      validationSchema={Validator}
+      onSubmit={onSubmit}
+      submitTrackerId="generate_rs_tidsid"
+      onClickResultRow={(results) => {
+        onSelectTarget(results.advance, results);
+      }}
```
zaksabeast, line 113: React re-renders when it detects a prop has changed:
```
1 === 1 // true: no change
'test' === 'test' // true: no change
10 === 11 // false: change
```

But here's the tricky bit: identical functions have different memory locations, which will always trigger changes:
```
const fn1 = () => 1;
const fn2 = () => 1;
fn1 === fn2; // false: change
```

The `RsTidSidGenerator` is a function that gets called when React tries to render it, and this line creates a new function every single time that happens.  React will always think `RngToolForm` needs to re-render.

`useCallback` is a hook that makes sure the component is memoized and will have the same memory location each time.

Let's pull this into a `useCallback`:
```tsx
const onClickResultRow = React.useCallback(
  (results: Result) => {
    onSelectTarget(results.advance, results);
  },
  [onSelectTarget],
);
```

We shouldn't reach for `useCallback` in every single component.  However forms are notorious for being huge performance vacuums, so it's good to be defensive here.

src/rngToolsUi/gen3/rstid/rstid.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const onSelectTarget = (advance: number, results) => {
+    handleTarget(advance, results);
+  };
+
+  const formatTime = (seconds: number): string => {
+    const hrs = Math.floor(seconds / 3600);
+    const mins = Math.floor((seconds % 3600) / 60);
+    const secs = Math.floor(seconds % 60);
+    return `${hrs}:${mins.toString().padStart(2, "0")}:${secs.toString().padStart(2, "0")}`;
+  };
```
zaksabeast, line 81: Similar to my comment below, this will create a new function every time this component is rendered.

Let's pull this outside the component since it doesn't use any values inside the component.

src/rngToolsUi/gen3/rstid/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  const onSelectTarget = (results: TargetResult) => {
+    isTargetSet(true);
+    setTarget(results);
+  };
+
+  return (
+    <Flex gap={32} vertical>
+      <Stepper onChange={onChange} current={current} items={items} />
+      <Flex vertical display={current < items.length - 1 ? "flex" : "none"}>
```
zaksabeast, line 69: Tiny nitpick: it's a bit odd to have the second tab show when `current === 1`, but have this one show when `current < items.length - 1`.

It's easier and more consistent to read `current === 0`:
```suggestion
      <Flex vertical display={current === 0 ? "flex" : "none"}>
```

src/rngToolsUi/gen3/rstid/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+    title: "Find Target TID",
+    //   subTitle: "00:00:05",
+    status: "process",
+    //   description: "Generate TID and possible SIDs.",
+  },
+  {
+    title: "RNG TID",
+    //   subTitle: "00:01:02",
+    status: "process",
+    //   description: "RNG for a shiny starter.",
```
zaksabeast, line 21: Let's remove or uncomment these.

src/rngToolsUi/gen3/rstid/stepper.tsx
```diff
@@ -0,0 +1,5 @@
+import { Steps, StepsProps } from "antd";
+
+export const Stepper = (props: StepsProps) => {
+  return <Steps type="navigation" size="small" {...props} />;
```
zaksabeast, line 4: Is there a reason this shouldn't be in `index.tsx`?  This entire file is basically just setting two additional props on top of whatever is used in the other file.

src/rngToolsUi/gen3/rstid/searcher.tsx
```diff
@@ -0,0 +1,116 @@
...
+  { title: "TID", dataIndex: "tid" },
+  { title: "SID", dataIndex: "sid" },
+  { title: "TSV", dataIndex: "tsv" },
+];
+
+const Validator = z.object({
+  offset: z.number().int().min(0),
+  initial_advances: z.number().int().min(0),
+  max_advances: z.number().int().min(0),
+  filter: IdFilterSchema,
```
zaksabeast, line 28: Since users will only ever filter by tid, we should probably just have a tid field.

In the validator we'll want:
```suggestion
  hit_tid: z.number().int().min(0).max(65535),
```

src/rngToolsUi/gen3/rstid/searcher.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+const initialValues: FormState = {
+  initial_advances: 0,
+  max_advances: 1000,
+  offset: 0,
+  filter: {
+    type: "tid",
+    value0: 0,
+    value1: null,
+  },
```
zaksabeast, line 41:
```suggestion
  hit_tid: 0,
```

src/rngToolsUi/gen3/rstid/searcher.tsx
```diff
@@ -0,0 +1,116 @@
...
+  {
+    label: "Max Advances",
+    input: (
+      <FormikNumberInput<FormState> name="max_advances" numType="decimal" />
+    ),
+  },
+  {
+    label: "Filter",
+    input: <FormikIdFilter<FormState> name="filter" optional />,
```
zaksabeast, line 59:
```suggestion
    label: "Hit TID",
    input: <FormikNumberInput<FormState> name="hit_tid" numType="decimal" />,
```

src/rngToolsUi/gen3/rstid/searcher.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const onSubmit = React.useCallback<RngToolSubmit<FormState>>(
+    async (opts) => {
+      const results = await rngTools.gen3_tidsid_states({
+        offset: opts.offset,
+        initial_advances: opts.initial_advances,
+        max_advances: opts.max_advances,
+        version_options: {
+          Rs: "DeadBattery",
+        },
+        filter: denormalizeIdFilter(opts.filter),
```
zaksabeast, line 83:
```suggestion
        filter: denormalizeIdFilter({
          type: "tid",
          value0: opts.hit_tid,
          value1: null,
        }),
```

src/rngToolsUi/gen3/rstid/searcher.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  return (
+    <RngToolForm<FormState, Result>
+      fields={fields}
+      columns={columns}
+      results={results}
+      initialValues={initialValues}
+      validationSchema={Validator}
+      onSubmit={onSubmit}
+      submitTrackerId="generate_rs_tidsid"
```
zaksabeast, line 111: This is a duplicate id from another file:
```suggestion
      submitTrackerId="search_rs_tidsid"
```

src/rngToolsUi/gen3/rstid/rstid.tsx
```diff
@@ -0,0 +1,120 @@
...
+      max_advances: opts.max_advances,
+      version_options: {
+        Rs: "DeadBattery",
+      },
+      filter: denormalizeIdFilter(opts.filter),
+    });
+
+    const updatedResults = results.map((r) => ({
+      ...r,
+      time: formatTime(r.advance / 30),
```
zaksabeast, line 94: I'm pretty sure this needs to be `59.7275`.

zaksabeast (changes requested):

ShinySylveon04:
In this PR I added my own stepper components, but now the main branch has its own stepper component separate from this one. How should I go about reconciling the two?

zaksabeast (approved):

PR #295, opened 2025-05-14
Update Page Descriptions

ShinySylveon04:
Pages I've updated so far:
- Mystic Timer
- SM/USUM Island Scan Pokemon List

Are there any others that could use an update?

### DIFF
--- a/guides/Tools and Emulators/Timer.mdx
+++ b/guides/Tools and Emulators/Timer.mdx
@@ -1,6 +1,6 @@
 ---
-title: "Mystic Timer"
-description: "A mobile-friendly alternative to Eon Timer, perfect for Gen 3, Gen 4, and Gen 7 RNG."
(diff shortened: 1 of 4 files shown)


guides/Tools and Emulators/Timer.mdx
```diff
@@ -1,6 +1,7 @@
 ---
-title: "Mystic Timer"
-description: "A mobile-friendly alternative to Eon Timer, perfect for Gen 3, Gen 4, and Gen 7 RNG."
+title: "Mystic Timer - Online Pokémon RNG Timer for Gen 3, 4, and 7"
+navDrawerTitle: Mystic Timer
+description: "Mystic Timer is a mobile-friendly alternative to Eon Timer. Perfect for Gen 3, 4, and 7 Pokémon RNG on any device."
```
zaksabeast, line 4: I like this, but should we add a Mac/Linux reference?

zaksabeast (approved):

PR #310, opened 2025-05-18
Gen3 Glossary 

RainingChain:


src/assets/glossary_gen3.json
```diff
@@ -0,0 +1,247 @@
...
+    "name": "PKHex",
+    "url": "https://github.com/kwsch/PKHeX",
+    "importance": 2,
+    "desc": "Pokémon save editor."
+  },
+  {
+    "name": "Trainer card flip",
+    "importance": 2,
+    "desc": "Technique to reduce noise in certain cases."
+  },
```
zaksabeast, line 220: Since there hasn't been any evidence for this, we can remove it.

src/rngToolsUi/shared/glossary.tsx
```diff
@@ -0,0 +1,104 @@
...
+    },
+  },
+];
+
+export const Glossary = ({ jsonUrl }: Props) => {
+  const [terms, setTerms] = React.useState<Term[]>([]);
+
+  React.useEffect(() => {
+    fetch(jsonUrl).then(async (res) => {
+      const termsInJson: TermInJson[] = await res.json();
```
zaksabeast, line 79: What's the rationale for providing a json url and fetching is instead of a TS file that can be type checked?

I think I know, but want to be sure and don't want to make assumptions.

notes.txt
zaksabeast, line 1: Was this meant to be checked in?

src/assets/glossary_gen3.json
zaksabeast, line 1: Someone on reddit asked what "retail" meant just yesterday.  Not a requirement, but should we add retail vs. emulator too?

zaksabeast (commented):
Overall, this is incredible!

I just left a few questions.

src/rngToolsUi/shared/glossary.tsx
RainingChain: Either way works. I personally prefer when raw data is in data format like JSON. I can change it if you want.

notes.txt
RainingChain: No

RainingChain:
@zaksabeast Made some changes. You can review again.

src/rngToolsUi/shared/glossary.tsx
zaksabeast: Ah, I thought you might have been trying to code split the bundle.

I lean towards a TS file for better type errors.

If we have 5 glossaries in the future and somehow one of the json files is missing a property or we edit the `TermInJson` type, the type error will be on `termsInJson`, which makes it difficult to figure out where the issue exists.

If each glossary is `export const glossary: TermsInJson[] = { /* stuff */ };`, we'll know the exact file and object in the array causing any type errors.

I'm happy to merge the PR as is now and make this change myself if you'd like.

src/assets/glossary_gen3.json
zaksabeast: Yeah, that's fair.

Thinking on it more, I shouldn't have asked about this.  It needs more thought.

Here's some historical context around how terms have evolved over time in the community.

Nearly 10 years ago, "retail" and "emulator" referred to whether a game was played on physical hardware or emulated hardware.  Most people who played with emulators used lua scripts, so there wasn't a distinction between the two.

During the 3ds era, PCalc (PokeReader's predecessor) was used on retail hardware, so it was "retail RNG".  The distinction was pcalc vs. not-pcalc.  This is why some of our PokeReader slugs have the word "retail" in it.

Eventually people started using "retail" to mean "not pcalc", 3ds emulators added scripting support, and "cfw" was used interchangeably with "pcalc", so we had "retail", "cfw", and "emulator".

Not too long ago PokeReader started working on emulators, so recently there's been little difference between "emulator" and "cfw" for the 3ds.  That's why the 3ds pages are a bit ambiguous.

Nowadays people are more interested in RNGing on emulators without lua scripts, using unofficial hardware, using custom firmware on a 3ds without Reader, etc.

At some point people started using arbitrary code execution and external hardware to access and modify game state, like the player's SID and initial seed, and called it "retail" still.  So "retail" doesn't even mean having no access to game state.

At this point none of the terms make any sense.

"Tool assisted" was introduced as a term to the RNG community just a few months ago.  It's a very new term, and doesn't seem widely adopted yet.

I'd like to push the community to use "Tool assisted" to mean lua scripts, Reader, external hardware, etc.  "Assisted RNG" might also be good so it includes ACE.  I'd also like to get away from using "Retail" as a term for similar reasons, though I'm not sure how easy that would be.

This needs more thought to have consistency, and the site should be organized around those terms.  That will create a better environment for adoption.

Let's remove those terms for now.  If you'd like, I'm happy to merge the PR as is and remove those terms myself so you don't have to.

@RainingChain Sound good?  Any thoughts on this?

src/rngToolsUi/shared/glossary.tsx
RainingChain: Well, I did write it so the json file is only fetched if the page actually needs it. However, idk how it's actually bundled in the end. With the way I coded it, there will be a compile-time error in case there's a bad attribute, except for "importance" field. 
I'll let you can change it for `export const glossary: TermsInJson[]` if you want.

src/assets/glossary_gen3.json
RainingChain: I don't really like the term "Tool assisted" because tool is too vague. PokeFinder is a tool, but using it doesn't make it "tool assisted".  I'll try thinking about something.

I agree to remove the terms. I'll let you do it.

src/rngToolsUi/shared/glossary.tsx
zaksabeast: Gotcha - it wouldn't be bundled the way you wrote it both times, but using `await import("file")` gives us some additional benefits.

You're right about getting a type error currently, but here's the difference between the two methods:
| Current implementation | As TS file |
| - | - |
| Screenshot 2025-05-20 at 4 19 38 PM | Screenshot 2025-05-20 at 4 21 38 PM |

With the current method, only `termsInJson` will get an error.  We don't know which file is incorrect, nor which object in the file is incorrect.

Typing each glossary as `TermsInJson[]` will tell us the exact file and object that's incorrect.

src/assets/glossary_gen3.json
zaksabeast: That's good point.  I've had similar thoughts, but figured I was overthinking it.

If you're having the same thoughts, it's a good sign to keep iterating.

zaksabeast (approved):
Thank you!

PR #312, opened 2025-05-20
RS Live Battery Static RNG Chinese translation

HaKu76:


zaksabeast (approved):
Thank you!  These translations have been amazing 😄

PR #316, opened 2025-05-24
Code for gen 4 static

sorendog:
adds code for tool that mimics stationary/static encounter in gen 4 using method 1, j and k.

currently functional, but going to attempt to add lead ability function calls for method j and k before calling this finished.

### DIFF
--- a/rng_tools/src/generators/gen4/stationary/generator.rs
+++ b/rng_tools/src/generators/gen4/stationary/generator.rs
@@ -1,0 +1,829 @@
+use crate::Ivs;
+use crate::rng::Rng;
+use crate::rng::StateIterator;
+use crate::rng::lcrng::Pokerng;
+use crate::{AbilityType, Gender, Nature, PkmFilter, PkmState, gen3_shiny};
+
+use super::GameVersion;
+use super::StaticEncounterId;
+use super::dpt_method_jk;
+use super::hgss_method_jk;
+
+//hi
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct Gen4SOpts {
+    pub tid: u16,
+    pub sid: u16,
+    pub initial_advances: usize,
+    pub max_advances: usize,
+    pub filter: PkmFilter,
+    pub game: Option<GameVersion>,
+    pub encounter: StaticEncounterId,
+}
+
+#[derive(Debug, PartialEq)]
+pub struct Gen4SPokemon {
+    pub pid: u32,
+    pub shiny: bool,
+    pub ability: AbilityType,
+    pub gender: Gender,
+    pub ivs: Ivs,
+    pub nature: Nature,
+    pub advance: usize,
+}
+
+impl PkmState for Gen4SPokemon {
+    fn ability(&self) -> AbilityType {
+        self.ability
+    }
+
+    fn gender(&self) -> Gender {
+        self.gender
+    }
+
+    fn ivs(&self) -> &Ivs {
+        &self.ivs
+    }
+
+    fn nature(&self) -> Nature {
+        self.nature
+    }
+
+    fn shiny(&self) -> bool {
+        self.shiny
+    }
+}
+
+pub fn generate_gen4_static(rng: &mut Pokerng, settings: Gen4SOpts) -> Option<Gen4SPokemon> {
+    let pid_low = rng.rand::<u16>() as u32;
+    let pid_high = rng.rand::<u16>() as u32;
+    let pid = (pid_high << 16) | pid_low;
+
+    let iv1 = rng.rand::<u16>();
+    let iv2 = rng.rand::<u16>();
+    let ivs = Ivs::new_g3(iv1, iv2);
+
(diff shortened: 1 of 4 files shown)


sorendog:
a review would be helpful now!

rng_tools/src/generators/gen4/stationary/generator.rs
```diff
@@ -0,0 +1,1416 @@
...
+                },
+                nature: Nature::Docile,
+                advance: 10,
+            },
+        ];
+        let result = filter_4static(options, seed);
+        assert_eq!(result, expected_results);
+    }
+    #[cfg(test)]
+    mod test {
```
zaksabeast, line 977: Instead of nesting `test::test` and having `synch` in the test name, it might be slightly cleaner to rename this test module to `synch` and rename the test names to their methods.

That reads slightly cleaner at a glance:
Screenshot 2025-05-27 at 8 35 21 AM

rng_tools/src/generators/gen4/stationary/generator.rs
```diff
@@ -0,0 +1,1416 @@
...
+                    spa: 9,
+                    spd: 6,
+                    spe: 29,
+                },
+                nature: Nature::Bold,
+                advance: 10,
+            },
+        ];
+        let result = filter_4static(options, seed);
+        assert_eq!(result, expected_results);
```
zaksabeast, line 539: One last thought: let's use `assert_list_eq` in these tests.

rng_tools/src/generators/gen4/stationary/methods.rs
```diff
@@ -0,0 +1,177 @@
...
+    Suicune,
+    Voltorb,
+    Snorlax,
+}
+
+impl StaticEncounterId {
+    pub fn species(self) -> Species {
+        match self {
+            StaticEncounterId::Chikorita => Species::Chikorita,
+            StaticEncounterId::Cyndaquil => Species::Cyndaquil,
```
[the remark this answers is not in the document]
sorendog: This is really good to know. I will keep this in mind next time it could be helpful!

sorendog:
Alright changes requested have been added, so review requested!

rng_tools/src/generators/gen4/stationary/generator.rs
```diff
@@ -0,0 +1,1847 @@
...
+                            shiny: gen3_shiny(pid, settings.tid, settings.sid),
+                            ability: AbilityType::from_gen3_pid(pid),
+                            gender,
+                            ivs,
+                            nature,
+                            advance: 0,
+                        });
+                    };
+                }
+                break;
```
zaksabeast, line 117: Hmm - something feels a bit off here.  We have a loop, but we break on the first loop, so it never goes past the first iteration.

Do you have a reference link from PokeFinder of where this logic is coming from?

rng_tools/src/generators/gen4/stationary/generator.rs
```diff
@@ -0,0 +1,1847 @@
...
+    fn nature(&self) -> Nature {
+        self.nature
+    }
+
+    fn shiny(&self) -> bool {
+        self.shiny
+    }
+}
+
+pub fn generate_gen4_static(rng: &mut Pokerng, settings: Gen4StaticOpts) -> Option<Gen4SPokemon> {
```
zaksabeast, line 58: Let's rename `settings` to `opts` for consistency with other functions.

rng_tools/src/generators/gen4/stationary/generator.rs
```diff
@@ -0,0 +1,1847 @@
...
+        ability: AbilityType::from_gen3_pid(pid),
+        gender: settings.encounter.species().gender_from_pid(pid),
+        ivs,
+        nature: Nature::from_pid(pid),
+        advance: 0,
+    };
+    Some(pkm)
+}
+
+pub fn generate_gen4_static_k(rng: &mut Pokerng, settings: Gen4StaticOpts) -> Option<Gen4SPokemon> {
```
zaksabeast, line 79: Let's update all functions that use `settings: Gen4StaticOpts` to `settings: &Gen4StaticOpts`.  This is more efficient since we don't need to copy data each function.

zaksabeast (changes requested):
A few more small tweaks and we should be good!

rng_tools/src/generators/gen4/stationary/generator.rs
sorendog: i can't find the exact part i was referencing with that code, i think this was left over from my attempt to do what pokefinder does. i had a bit where i thought the ushort calls that have a number in parentheses meant they were making multiple calls, like this line: https://github.com/Admiral-Fish/PokeFinder/blob/3cd27951484dd83a331079827d2b3ff969d92165/Source/Core/Gen4/Searchers/StaticSearcher4.cpp#L202

but realized that isn't what they are doing basically.

tldr: i will get this loop removed.

sorendog:
changes made - hopefully didn't miss anything.

PR #319, opened 2025-05-28
Gen 4 wild

sorendog:
Eventually this will be a working gen 4 wild tool. Very Early stages atm but figured I would start the PR now (also using this to practice using multiple branches to manage my code instead of having every branch on my computer lol...)


sorendog:
Alright everything checked over - This is ready for review.

The only note I have is, Maybe I should change the generatedpokemon struct name to something else? I feel like I used that name somewhere else and i am unsure if that would cause any problems even though vs code did not point any out.

rng_tools/src/generators/gen4/gen4wild/generator.rs
```diff
@@ -0,0 +1,2186 @@
...
+    pub fn get_slot(&self, slot_id: u8) -> Option<&EncounterslotIDs> {
+        self.encounter_slots
+            .iter()
+            .find(|slot| slot.slot_id == slot_id)
+    }
+}
+#[derive(Debug, Clone, PartialEq)]
+pub struct EncounterslotIDs {
+    pub slot_id: u8,
+    pub pokemon_id: Species,
```
zaksabeast: Let's rename this to `species` for consistency with other parts of the codebase.

rng_tools/src/generators/gen4/gen4wild/generator.rs
```diff
@@ -0,0 +1,2186 @@
...
+impl RouteData {
+    pub fn get_slot(&self, slot_id: u8) -> Option<&EncounterslotIDs> {
+        self.encounter_slots
+            .iter()
+            .find(|slot| slot.slot_id == slot_id)
+    }
+}
+#[derive(Debug, Clone, PartialEq)]
+pub struct EncounterslotIDs {
+    pub slot_id: u8,
```
zaksabeast: I wonder if we should rename `EncounterSlotIds` to `Encounter` since this struct describes an encounter (the slot, species, and min/max levels).

Similarly, renaming `slot_id` to `slot` would match `EncounterSlot`, which is the 0-11 value we put here.

zaksabeast: If we rename `slot_id` to `slot` and give it the `EncounterSlot` type, we can also update `get_slot` to use `EncounterSlot`:
```suggestion
    pub fn get_slot(&self, encounter_slot: EncounterSlot) -> Option<&EncounterslotIDs> {
        self.encounter_slots
            .iter()
            .find(|slot| slot.slot == encounter_slot)
    }
}
#[derive(Debug, Clone, PartialEq)]
pub struct EncounterslotIDs {
    pub slot: EncounterSlot,
```

This way we can remove the new `slot_id` method from `EncounterSlot`.

All of this makes sure we can't accidentally use an invalid number where we expect an `EncounterSlot`.

rng_tools/src/generators/gen4/gen4wild/generator.rs
```diff
@@ -0,0 +1,2186 @@
+use crate::GenderRatio;
+use crate::Ivs;
+use crate::Species;
+use crate::gen3::EncounterSlot;
```
zaksabeast: If we're using this in both gen 3 and 4, we should move it to a shared location multiple gens can reference.

Perhaps `crate::encounter::EncounterSlot` for now?

rng_tools/src/generators/gen4/gen4wild/generator.rs
```diff
@@ -0,0 +1,2186 @@
...
+
+pub fn gen4_method_j(
+    rng: &mut Pokerng,
+    opts: &Gen4SWildOpts,
+    route: &RouteData,
+) -> Option<GeneratedPokemon> {
+    let encounter_rand = ((rng.rand::<u32>() >> 16) / 656) as u8;
+    let encounter_slot = EncounterSlot::from_rand(encounter_rand);
+    let slot = route.get_slot(encounter_slot.slot_id())?;
+
```
zaksabeast: It looks like right around here there's logic to get `battleAdvances` in PokeFinder and it seems to impact Diamond and Pearl games: https://github.com/Admiral-Fish/PokeFinder/blob/150aea364ade4ae17f81ad70ccca76a1c64971b1/Source/Core/Gen4/Generators/WildGenerator4.cpp#L123

Is there a reason this isn't needed here at the moment?

rng_tools/src/generators/gen4/gen4wild/generator.rs
```diff
@@ -0,0 +1,2186 @@
...
+) -> Option<GeneratedPokemon> {
+    match opts.game {
+        GameVersion::Diamond | GameVersion::Pearl | GameVersion::Platinum => {
+            gen4_method_j(rng, opts, route)
+        }
+        GameVersion::HeartGold | GameVersion::SoulSilver => gen4_method_k(rng, opts, route),
+    }
+}
+
+pub fn filter_4swild(opts: &Gen4SWildOpts, seed: u32, route: &RouteData) -> Vec<GeneratedPokemon> {
```
zaksabeast: Lets rename this to `search_wild4` and `Wild3SearcherOpts`.

Wile it is filtering, filtering is only one part of the entire search process.

rng_tools/src/generators/gen4/gen4wild/generator.rs
```diff
@@ -0,0 +1,2186 @@
...
+pub enum EncounterType {
+    Grass,
+    Cave,
+    Surf,
+    OldRod,
+    GoodRod,
+    SuperRod,
+}
+
+pub fn gen4_method_j(
```
zaksabeast: Let's remove `pub fn` from anything that isn't `filter_4swild`.

Nothing outside of these files should call these functions, so we don't want to expose them.

rng_tools/src/generators/gen4/gen4wild/generator.rs
```diff
@@ -0,0 +1,2186 @@
...
+    rng: &mut Pokerng,
+    opts: &Gen4SWildOpts,
+    route: &RouteData,
+) -> Option<GeneratedPokemon> {
```
zaksabeast: Let's update this to return a `GeneratedPokemon` - there shouldn't be a time we try to generate a Pokemon and return nothing.

It looks like this might be from `route.get_slot` returning an `Option`.

We can solve that by passing an `Encounter` to this function instead of a route that we use to find an encounter.

After that, we should have no more need for an `Option`.

Let's do the same with Method K as well.

rng_tools/src/generators/gen4/gen4wild/generator.rs
```diff
@@ -0,0 +1,2186 @@
...
+        gender: slot.pokemon_id.gender_from_pid(pid),
+        encounter_slot,
+        ivs,
+        nature: Nature::from_pid(pid),
+        advance: 0,
+    };
+    Some(pkm)
+}
+
+pub fn generate_4wild(
```
zaksabeast: Tiny nitpick: let's rename this to `generate_wild4` - we tend to put the generation after the Pokemon type.

rng_tools/src/generators/gen4/gen4wild/generator.rs
```diff
@@ -0,0 +1,2186 @@
...
+    GoodRod,
+    SuperRod,
+}
+
+pub fn gen4_method_j(
+    rng: &mut Pokerng,
+    opts: &Gen4SWildOpts,
+    route: &RouteData,
+) -> Option<GeneratedPokemon> {
+    let encounter_rand = ((rng.rand::<u32>() >> 16) / 656) as u8;
```
zaksabeast: It looks like you're using `rng.rand::<u32>() >> 16` and `rng.rand::<u16>()`.

These are the same - only one needs to be used.  I like `rng.rand::<u16>()` since its shorter and shows the intent better.

rng_tools/src/generators/gen4/gen4wild/generator.rs
sorendog: When i looked into battle advances, it looked like any call to it was so the number could be advanced everytime the rng is being called but the number isn't actually used for determining anything. Its just how many advances a battle being triggered will cause. I can see how that could be helpful for some, but since it isn't needed for determining stats or encounter slot, I haven't implemented it. I can add it if you think otherwise? Or if I'm wrong of course in my understanding.

zaksabeast: We can add it later if we find it's needed for a future RNG.  Having a solid base to start is good for now 👍

rng_tools/src/generators/gen3/wild/types.rs
```diff
@@ -46,6 +46,9 @@ impl EncounterSlot {
             (Slot11, 100),
         ]
     }
+    pub fn slot_id(&self) -> u8 {
+        *self as u8
+    }
```
zaksabeast: Since `EncounterSlot` now lives at `crate::EncounterSlot`, we can remove the `EncounterSlot` in this file since they're duplicates.

Running `cargo clippy` will show a few places where `crate::gen3::EncounterSlot` will need to be udpated to `crate::EncounterSlot`.

rng_tools/src/pkm/encounter.rs
```diff
@@ -0,0 +1,61 @@
...
+            (Slot7, 90),
+            (Slot8, 94),
+            (Slot9, 98),
+            (Slot10, 99),
+            (Slot11, 100),
+        ]
+    }
+    pub fn slot_id(&self) -> u8 {
+        *self as u8
+    }
```
zaksabeast: We can remove this since it's not being used at the moment.

zaksabeast (changes requested):
Two tiny clean up items and we should be good

zaksabeast (approved):
Thank you!

PR #322, opened 2025-05-29
How to Extract GBA Bios Chinese translation

HaKu76:


zaksabeast (commented):
@HaKu76 One quick question, otherwise this looks great!

PR #326, opened 2025-05-31
Wild3 Find Target Pokémon UI Tool (MVP)

RainingChain:
image

Other PRs will add improvements listed in wild3FindTarget.tsx


src/rngToolsUi/gen3/wild/wild3FindTarget.tsx
```diff
@@ -0,0 +1,434 @@
...
+          disabled={!filter_shiny}
+        />
+      ),
+    },
+    {
+      label: "Maps",
+      input: (
+        <FormikSelect<FormState, "maps">
+          name="maps"
+          options={[...toOptions(mapsWithSpecies)]}
```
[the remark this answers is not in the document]
RainingChain: No.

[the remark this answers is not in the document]
RainingChain: I guess that's better than what I have atm. Still, in the long term, if we plan on supporting translations, we will need a proper map names list.

src/rngToolsUi/gen3/wild/wild3FindTarget.tsx
```diff
@@ -0,0 +1,434 @@
...
+  return {
+    species: "Abra",
+    tid: 0,
+    sid: 0,
+    maps: [],
+    vanillaLead: true,
+    cuteCharmLeadGenders: [...cuteCharmGenders],
+    synchronizeLeadNatures: [...nature],
+    methods: ["Wild2"],
+    encounterTypes: [...gen3EncounterTypes],
```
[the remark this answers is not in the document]
SteveCookTU: Any objection to using `Object.freeze()` for the arrays and data we don't want changed?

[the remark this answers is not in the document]
zaksabeast: Yeah, `Object.freeze()` works.

We can also add `readonly` to the types:
```ts
const cuteCharmGenders = [
  "Male",
  "Female",
] as const satisfies readonly Gender[];
```

Screenshot 2025-05-31 at 9 53 27 PM

The main issue is if we cast the type to remove the `readonly`.  `Object.freeze` would error at runtime.  `readonly` by itself would still be functional for the user, unless it was a really bad cast.  Both aren't ideal, and as long as we don't cast anything we don't have to worry either way.

I like how `readonly` is a bit more terse, but don't have a super strong preference for which way we go.

src/rngToolsUi/gen3/wild/emeraldWildGameData.ts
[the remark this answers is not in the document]
zaksabeast: I responded in Discord since more people might see that thread.

But, yeah that sounds good to me.

src/rngToolsUi/gen3/wild/wild3FindTarget.tsx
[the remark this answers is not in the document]
RainingChain: @zaksabeast Object.freeze() and readonly are kinda the same solution (Object.freeze() returns a readonly array). readonly array requires that all users of the array must specifically permit readonly array. 

Forcing the form state to have readonly arrays just because its initial values are initialized from readonly arrays seem weird. The copy of the array seems justified to me.

Converting all arrays in `/types` to be readonly seems a good idea. That could be done in a separate PR.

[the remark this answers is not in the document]
zaksabeast: Yeah, that's what I was getting at originally - they're basically the same solution.  And a separate PR would be good to prevent scope inflation.

@SteveCookTU @RainingChain Actually I just did some testing and found out `as const satisfies readonly` does a deep comparison:
```tsx
const arr1 = [{ name: 'test'}] as const satisfies readonly { name: string }[];
// Type error
arr1[0].name = 'test2';

const arr2: readonly { name: string }[] = [{ name: 'test'}];
// No type error
arr2[0].name = 'test2';

const arr3 = Object.freeze([{ name: 'test'}]);
// No type error
arr3[0].name = 'test2';
```

If we're concerned about mutability, we should use `as const satisfies readonly Type` for deep coverage.

@RainingChain If the intention behind copying the array is to make sure we can't mutate it, ensuring it's readonly, and we want to apply `readonly` to `/types`, is there a reason to not adopt that standard everywhere?

Spreading also only makes a shallow clone.  If we're concerned about mutability, that won't give us as much protection.

Any thoughts from you two about this?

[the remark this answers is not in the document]
RainingChain: @zaksabeast The intention behind copying the array is to avoid marking the props of the form state as readonly. (Assuming all type arrays become readonly in the near future.)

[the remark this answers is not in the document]
zaksabeast: I was thinking we might update the validator to ensure types are readonly, but that seems more of a hassle than its worth.

This looks good to me!

scripts/wild3/generateWildGameData.ts
```diff
@@ -0,0 +1,147 @@
...
+
+pub struct Wild3EncounterTable {
+    pub map_id:String,
+    pub encounter_rate:u8,
+    pub slots:Vec<Wild3EncounterSlotInfo>,
+}
+*/
+
+  await writeFile(
+    "./src/rngToolsUi/gen3/wild/emerald_wild3_game_data.ts",
```
zaksabeast: Should we put this in `__generated__`?  The message at the top saying `DO NOT MANUALLY EDIT THIS FILE` is great, although I wonder if the dedicated folder would also be good so we can keep all generated data together.

scripts/wild3/generateWildGameData.ts
```diff
@@ -0,0 +1,147 @@
+import type {
+  Species,
+  EncounterSlot,
+  Wild3EncounterSlotInfo,
+  Wild3EncounterTable,
+} from "../../src/rngTools";
+import { species as allSpecies, species } from "../../src/types/species";
+import { genderRatioBySpecies } from "../../src/types/gender";
+import { encounterSlots } from "../../src/types/encounterSlots";
```
zaksabeast: Looks like we have some unused imports:
```suggestion
import type { Species } from "../../src/rngTools";
import { species as allSpecies } from "../../src/types/species";
import { genderRatioBySpecies } from "../../src/types/gender";
```

scripts/wild3/generateWildGameData.ts
```diff
@@ -0,0 +1,147 @@
...
+        map?: string;
+        base_label?: string;
+        land_mons?: {
+          mons: {
+            min_level: number;
+            max_level: number;
+            species: string;
+          }[];
+        };
+      };
```
zaksabeast: Let's use a zod schema to validate this:
```ts
const EncounterSchema = z.object({
  map: z.string().optional(),
  base_label: z.string().optional(),
  land_mons: z.object({
    mons: z
      .object({
        min_level: z.number(),
        max_level: z.number(),
        species: z.string(),
      })
      .array(),
  }).optional(),
});

const enc = EncounterSchema.parse(raw_enc);
```

That way if something changes at some point in the future, we'll get an error when running this script to let us know there's a problem.

scripts/wild3/generateWildGameData.ts
```diff
@@ -0,0 +1,147 @@
...
+    pub is_electric_type:bool,
+    pub is_steel_type:bool,
+}
+
+pub struct Wild3EncounterTable {
+    pub map_id:String,
+    pub encounter_rate:u8,
+    pub slots:Vec<Wild3EncounterSlotInfo>,
+}
+*/
```
zaksabeast: Is there a reason to keep these comments?  The logic and types seem to cover the info here.

scripts/wild3/generateWildGameData.ts
```diff
@@ -0,0 +1,147 @@
...
+                "min_level": 2,
+                "max_level": 2,
+                "species": "SPECIES_WURMPLE"
+              },
+            ]
+          }
+        },
+
+
+*/
```
zaksabeast: Is there a reason to keep these comments?

scripts/wild3/generateWildGameData.ts
```diff
@@ -0,0 +1,147 @@
...
+// DO NOT MANUALLY EDIT THIS FILE
+// This file was generated by running "bun run build:wild3".
+
+import type { Wild3GameDataJSON } from "./wild3GameData";
+
+const wild3_game_data: Wild3GameDataJSON = ${JSON.stringify(jsonData)};
+
+export default wild3_game_data;
+`;
+  await fs.writeFile(destinationFile, str);
```
zaksabeast: Let's run prettier on the output to make sure it doesn't accidentally cause format errors at some point:
```ts
import prettier from "prettier";

prettier.format(str, { parser: "typescript" })
```

scripts/wild3/pokemonTypes.ts
```diff
@@ -0,0 +1,421 @@
...
+  Regice: ["Ice"],
+  Registeel: ["Steel"],
+  Latias: ["Dragon", "Psychic"],
+  Latios: ["Dragon", "Psychic"],
+  Kyogre: ["Water"],
+  Groudon: ["Ground"],
+  Rayquaza: ["Dragon", "Flying"],
+  Jirachi: ["Steel", "Psychic"],
+  Deoxys: ["Psychic"],
+} as Record<string, POKEMON_TYPE[]>;
```
zaksabeast: Let's use `satisfies` since the type error is easier to spot at a glance:
```ts
} as const satisfies Record<string, POKEMON_TYPE[]>;
```

Screenshot 2025-06-02 at 8 33 53 AM

scripts/wild3/pokemonTypes.ts
```diff
@@ -0,0 +1,421 @@
...
+
+export const doesSpeciesHaveType = (
+  _gen: 3,
+  species: Species,
+  type: POKEMON_TYPE,
+) => {
+  const types = gen3TypesBySpecies[species];
+  if (!types) throw new Error(`Unsupported species ${species}`);
+  return types.includes(type);
+};
```
zaksabeast, line 419: This will fail for some species at the moment:
```sh
% bun repl
> import { doesSpeciesHaveType } from './scripts/wild3/pokemonTypes'
undefined
> doesSpeciesHaveType(3, "FarfetchD", "Normal")
412 |   _gen: 3,
413 |   species: Species,
414 |   type: POKEMON_TYPE,
415 | ) => {
416 |   const types = gen3TypesBySpecies[species];
417 |   if (!types) throw new Error(`Unsupported species ${species}`);
                          ^
error: Unsupported species FarfetchD
```

If we want to use `Species` to index `gen3TypesBySpecies`, we should give it that constraint:
```ts
as const satisfies Partial<Record<Species, POKEMON_TYPE[]>>
```

`Partial` lets use specify only the fields we want, so we don't need types for every species.

Doing this shows a few places where `gen3TypesBySpecies` doesn't match the expected value.

scripts/wild3/generateWildGameData.ts
```diff
@@ -0,0 +1,147 @@
...
+  };
+  const jsonNameToSpecies = (speciesName: string): Species | undefined => {
+    return formattedSpeciesToSpecies.get(formatSpecies(speciesName));
+  };
+
+  const formattedSpeciesToSpecies = new Map<string, Species>();
+
+  allSpecies.forEach((species) => {
+    formattedSpeciesToSpecies.set(formatSpecies(species), species);
+  });
```
zaksabeast: Given the species issue in the other file, we might want to be a little more defensive here and error when a species doesn't exist instead of returning `undefined`:
```ts
  const SpeciesSchema = z.enum(allSpecies);
  const jsonNameToSpecies = (speciesName: string): Species => {
    return SpeciesSchema.parse(upperFirst(formatSpecies(speciesName)));
  };
```

src/rngToolsUi/gen3/wild/wild3GameData.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export type Wild3GameData = {
+  encounter_tables: Wild3EncounterTable[];
+  species: Species[];
+  maps: string[];
+  speciesToEncounterSlots: Map<Species, Map<string, EncounterSlot[]>>;
+};
+
+const createSortedArrayWithoutDuplicates = <T extends string>(array: T[]) => {
+  return Array.from(new Set(array)).sort((lhs, rhs) => lhs.localeCompare(rhs));
```
zaksabeast: Not necessary to change in this PR, but in case its helpful in the future, `lodash-es` exports `uniq` ([docs here](https://lodash.com/docs/4.17.15#uniq)), which creates a new duplicate-free array:
```ts
import { uniq } from "lodash-es";
console.log(uniq([1, 2, 2, 3, 4, 4, 5]));
// [1, 2, 3, 4, 5]
```

`uniqBy` can be used to apply some logic on each item, such as pulling an id from an object.

scripts/wild3/pokemonTypes.ts
RainingChain: Yeah, that's what I wanted to do at first but didn't think of using Partial.

scripts/wild3/generateWildGameData.ts
RainingChain: `upperFirst` won't return Species , because some Species have special formats like `MrMime` and `FarfetchD`. I can add a throw directly in `jsonNameToSpecies` instead of in the caller.

PR #328, opened 2025-05-31
Wild3 add cute charm support

RainingChain:


rng_tools/src/generators/gen3/wild/generator.rs
```diff
@@ -45,13 +45,29 @@ pub fn generate_gen3_wild(
...
+                })
+            } else {
+                None
+            }
+        } else {
+            None
+        }
+    } else {
+        None
+    };
```
zaksabeast: What are thoughts on something like this to trim down the `None`s:
```rs
let generated_mon_required_gender = match (opts.gender_ratio.has_multiple_genders(), opts.lead)
{
    (true, Some(Gen3Lead::CuteCharm(gender))) if rng.rand::<u16>() % 3 != 0 => {
        Some(if gender == Gender::Female {
            Gender::Male
        } else {
            Gender::Female
        })
    }
    _ => None,
};
```

zaksabeast (approved):
One quick question to possibly tidy something up, but this is good as-is. 👍

zaksabeast (approved):
Nice!

PR #330, opened 2025-06-04
Add concurrency proposal

zaksabeast:
View the pretty markdown [here](https://github.com/zaksabeast/PokemonRNGGuides/blob/64d51ca5360ab34d7cc798baea799c9b896d57c5/proposal/concurrency.md).

### DIFF
--- a/proposal/concurrency.md
+++ b/proposal/concurrency.md
@@ -22,7 +22,7 @@ const results = await multiWorkerRngTools.search_dppt_ids(opts);
 
 If this works well, we can remove the old `rngTools` and rename `multiWorkerRngTools` back to `rngTools`.
 
-We'll need to be careful not to spawn too many workers at once, since that can cause performance issues. We should keep it under 4 active workers at a time.
+We'll need to be careful not to spawn too many workers at once, since that can cause performance issues. Concurrency limiting tools will be added in the next section of this document.
 
 ## Batching and concurrency
 
@@ -48,26 +48,45 @@ const mult2 = async (num: number) => [num * 2];
 To make use a batchable function with chunked args, we'll add a new hook called `useBatchedTool`. It takes a batchable function and returns a function that accepts a list of the same input type.
 
 ```ts
-const [doMath] = useBatchedTool(mult2);
+const { run: doMath } = useBatchedTool(mult2);
 const result = await doMath([1, 2, 3]);
 console.log(result); // [2, 4, 6]
 ```
 
-`useBatchedTool` returns a tuple, `[runFn, meta, cancelFn]`, where meta includes progress, loading, etc.
+`useBatchedTool` returns an object:
+
+```ts
+type BatchedTool<Arg, Ret, MappedRet = Ret> {
+  run: (args: Arg[]) => Ret[]; // Process the chunked args
+  cancel: () => void; // Cancel the operation
+  data: MappedRet[]; // Current processed data
+  loading: boolean; // If is processing processing
+  progressPercent: number; // processedArgs / totalArgs * 100
+  error: Error | null; // An error caught while process a chunk
+}
+```
 
 ### Mapping results
 
 We often need to add additional data to results, such as the offset between a result and target advance.
 
-`useBatchedTool` should support a second argument: a mapping function that transforms each result item.
+`useBatchedTool` should support a second argument with options to transform the results:
+
+```ts
+type BatchedToolOptions<Result, MappedResult = Result> {
+  map?: (res: Result) => MappedResult;
+  sortBy?: (res: MappedResult) => MappedResult;
+}
+```
 
 In this example we're multiplying the number as before, but transforming the result into an object.
 
 ```ts
 const mapper = (num: number) => ({ num });
-const [doMath] = useBatchedTool(mult2, mapper);
+const sorter = (num: number) => 0 - num;
+const { run: doMath } = useBatchedTool(mult2, { map: mapper, sortBy: sorter });
 const result = await doMath([1, 2, 3]);
-console.log(result); // [{ num: 2 }, { num: 4 }, { num: 6 }]
+console.log(result); // [{ num: 6 }, { num: 4 }, { num: 2 }]
 ```
 
 ### Concurrency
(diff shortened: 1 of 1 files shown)


proposal/concurrency.md
```diff
@@ -0,0 +1,218 @@
...
+// Currently
+const results = await rngTools.search_dppt_ids(opts);
+
+// After
+const results = await multiWorkerRngTools.search_dppt_ids(opts);
+```
+
+If this works well, we can remove the old `rngTools` and rename `multiWorkerRngTools` back to `rngTools`.
+
+We'll need to be careful not to spawn too many workers at once, since that can cause performance issues. We should keep it under 4 active workers at a time.
```
RainingChain: I'd suggest using `window.navigator.hardwareConcurrency` and then check if there are real issues before fallbacking to just 4.
Alternatively, we could lower the number only on mobile.

proposal/concurrency.md
```diff
@@ -0,0 +1,218 @@
...
+
```
RainingChain, line 26: Here's the code I have for multiple workers with rust on my site:
https://pastebin.com/zr7zJNm9

proposal/concurrency.md
```diff
@@ -0,0 +1,218 @@
...
+To make use a batchable function with chunked args, we'll add a new hook called `useBatchedTool`. It takes a batchable function and returns a function that accepts a list of the same input type.
+
+```ts
+const [doMath] = useBatchedTool(mult2);
+const result = await doMath([1, 2, 3]);
+console.log(result); // [2, 4, 6]
+```
+
+`useBatchedTool` returns a tuple, `[runFn, meta, cancelFn]`, where meta includes progress, loading, etc.
```
RainingChain: I'd rather all have named variables.

const {runFn, cancelFn, ...}.
It's easier to add new stuff later.

proposal/concurrency.md
```diff
@@ -0,0 +1,218 @@
...
+Concurrency should default to 4, which is a safe number of workers to run at once.
+
+### Streamed results
+
+We should be able to access results as they're processed so users can see results as they're found.
+
+The example below shows the logging that occurs each render as new results are available. Notice how new results are logged over a few seconds.
+
+```ts
+const [doMath, { data }] = useBatchedTool(mult2);
```
RainingChain: Not quite clear to me how that works.
`data` is updated only when a worker ends? That seems too long.
`data` is updated for every result? That seems too quick.

How can we force refreshing the UI? I guess we need a callback.

proposal/concurrency.md
```diff
@@ -0,0 +1,218 @@
...
+
+### Loading, progress, and errors
+
+The hook should expose the current loading state, progress percentage, and any error that occurs. Notice how progress incremeents over time.
+
+```ts
+const [doMath, { loading, progressPercent, error }] = useBatchedTool(mult2);
+
+console.log(loading, progressPercent, error);
+// true 0 null
```
RainingChain, line 175: How often is it refreshed?

proposal/concurrency.md
```diff
@@ -0,0 +1,218 @@
...
+console.log(result); // [2, 4, 6]
+```
+
+`useBatchedTool` returns a tuple, `[runFn, meta, cancelFn]`, where meta includes progress, loading, etc.
+
+### Mapping results
+
+We often need to add additional data to results, such as the offset between a result and target advance.
+
+`useBatchedTool` should support a second argument: a mapping function that transforms each result item.
```
RainingChain: I'd also need a sort function (ideally something like binary insertion sort).

proposal/concurrency.md
zaksabeast: Great idea - let's go with `hardwareConcurrency`.

proposal/concurrency.md
zaksabeast: I'm good with that.  The main reason I went with a tuple is from it being the common pattern used in React.

It's also slightly nicer to give custom names to tuples `[myFunc]` than when destructuring `{ runFn: myFunc }`.

But I'm not attached to it in any way - named variables sounds good to me.

proposal/concurrency.md
zaksabeast: Makes sense!

proposal/concurrency.md
zaksabeast: Internally, `data` would be backed by a `useState`, so the UI will update any time there's new data from web workers.

If multiple `setStates` are called close enough together in certain contexts, React will batch their updates.

In the example below, there are six arguments (`[1, 2, 3, 4, 5, 6]`), which means six workers emitting six results, but `data` was re-rendered three times.

The max number of updates should be the max number of chunks.  Since chunking isn't prescribed by `useBatchedTool`, how often the UI re-renders is partially dependent on the caller.

proposal/concurrency.md
zaksabeast: It's reactive - updates will happen any time new results from workers are available.

RainingChain: @zaksabeast So it's the job of Rust to send the results in sub-chunks.

Like, I need to test 1 billion seeds.
I split it in 1000 chunks of 1 million seeds. Each worker processes 1 million seeds at a time. 
In the worker, at every 1000 results, I send the data to JS.

Is that what you were planning?

zaksabeast: Yeah, that's the idea.

`useBatchedTool` is agnostic to what does the calculating.  It's primarily scheduling jobs and collecting results as the jobs complete.

The batchable function doesn't need be aware it's searching a chunk.  In your example it would know how to take a range of seeds and filter for specific results.

The caller would break a large range into chunks, `useBatchedTool` would feed the chunks into the batchable function, the batchable function would return results as if it weren't searching a chunk, and `useBatchedTool` would collect the results as they're available.

zaksabeast:
@RainingChain Changes were made with your suggestions.

proposal/concurrency.md
zaksabeast: @RainingChain Thinking on this more, most users won't manually search through more than a few hundred results.  Showing users a few thousand results doesn't seem helpful.  Is there a time we'd want more than, say, 10,000 results from `useBatchedTool`?

Assuming there isn't a time we'd need more than that, and we cap our results to a few thousand, binary insertion might be a bit overkill - the default sort can handle a few thousand items in a few ms.

What are thoughts on using the default for now and improving performance if needed?

RainingChain: Yeah, I guess.

PR #331, opened 2025-06-04
TS for gen 4 stationary

sorendog:
Just the start so putting this as a draft - I might update or make a guide to go with this as well so can be useable in some fashion as soon as i am finished.

### DIFF
--- a/rng_tools/src/generators/gen4/stationary/generator.rs
+++ b/rng_tools/src/generators/gen4/stationary/generator.rs
@@ -1,16 +1,20 @@
-use crate::Ivs;
-use crate::rng::Rng;
-use crate::rng::StateIterator;
-use crate::rng::lcrng::Pokerng;
-use crate::{AbilityType, Gender, Nature, PkmFilter, PkmState, gen3_shiny};
+use wasm_bindgen::prelude::wasm_bindgen;
 
 use super::GameVersion;
 use super::LeadAbilities;
 use super::StaticEncounterId;
 use super::dpt_method_jk;
 use super::hgss_method_jk;
+use crate::Ivs;
+use crate::rng::Rng;
+use crate::rng::StateIterator;
+use crate::rng::lcrng::Pokerng;
+use crate::{AbilityType, Gender, Nature, PkmFilter, PkmState, gen3_shiny};
+use serde::{Deserialize, Serialize};
+use tsify_next::Tsify;
 
-#[derive(Debug, PartialEq, Clone)]
+#[derive(Debug, Clone, PartialEq, Tsify, Serialize, Deserialize, Copy)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
 pub struct Gen4StaticOpts {
     pub tid: u16,
     pub sid: u16,
@@ -22,7 +26,8 @@ pub struct Gen4StaticOpts {
     pub lead: Option<LeadAbilities>,
 }
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Tsify, Serialize, Deserialize, Copy)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
 pub struct Gen4SPokemon {
     pub pid: u32,
     pub shiny: bool,
@@ -288,6 +293,7 @@ pub fn generate_4statics(opts: &Gen4StaticOpts, rng: &mut Pokerng) -> Option<Gen
     }
 }
 
+#[wasm_bindgen]
 pub fn filter_4static(opts: &Gen4StaticOpts, seed: u32) -> Vec<Gen4SPokemon> {
     let base_rng = Pokerng::new(seed);
     StateIterator::new(base_rng)
--- a/rng_tools/src/generators/gen4/stationary/generator.rs
+++ b/rng_tools/src/generators/gen4/stationary/generator.rs
@@ -13,7 +13,7 @@ use crate::{AbilityType, Gender, Nature, PkmFilter, PkmState, gen3_shiny};
 use serde::{Deserialize, Serialize};
 use tsify_next::Tsify;
 
-#[derive(Debug, Clone, PartialEq, Tsify, Serialize, Deserialize, Copy)]
+#[derive(Debug, Clone, PartialEq, Tsify, Serialize, Deserialize)]
 #[tsify(into_wasm_abi, from_wasm_abi)]
 pub struct Gen4StaticOpts {
     pub tid: u16,
--- a/src/rngToolsUi/gen4/gen4stationary.tsx
+++ b/src/rngToolsUi/gen4/gen4stationary.tsx
@@ -1,0 +1,157 @@
+import { rngTools, Gen4StaticOpts } from "~/rngTools";
+import {
+  Field,
+  FormikNumberInput,
+  FormikSelect,
+  ResultColumn,
+  RngToolSubmit,
+} from "~/components";
+import { getPkmFilterFields, pkmFilterSchema } from "~/components/pkmFilter";
+import React from "react";
+import { maxIvs, minIvs } from "~/types/ivs";
+import {
+  flattenIvs,
+  FlattenIvs,
+  ivColumns,
+} from "~/rngToolsUi/shared/ivColumns";
+import {
+  GameVersion,
+  StaticEncounterId,
+} from "~/rngToolsUi/gen4/static/constants";
+import { z } from "zod";
+import { HexSchema } from "~/utils/number";
+
+type Result = FlattenIvs<Gen4StaticOpts>;
+
+const columns: ResultColumn<Result>[] = [
+  {
+    title: "PID",
+    dataIndex: "pid",
+    monospace: true,
+    render: (pid) => pid.toString(16).padStart(8, "0").toUpperCase(),
+  },
+  { title: "Nature", dataIndex: "nature" },
+  { title: "Ability", dataIndex: "ability" },
+  ...ivColumns,
+  {
+    title: "Shiny",
+    dataIndex: "shiny",
+    render: (shiny) => (shiny ? "Yes" : "No"),
+  },
+  { title: "Gender", dataIndex: "gender" },
+   {
+    title: "Advance",
+    dataIndex: "advance",
+    monospace: true,
+  },
+];
+
+const Validator = z
+  .object({
+    seed: HexSchema(0xffffffff),
+    tid: z.number().int().min(0).max(65535),
+    sid: z.number().int().min(0).max(65535),
+    initial_advances: z.number(),
+    max_advances: z.number(),
+    game: z.enum(GameVersion),
+    encounter: z.enum(StaticEncounterId),
+  })
+  .merge(pkmFilterSchema);
+
+type FormState = z.infer<typeof Validator>;
+
+const initialValues: FormState = {
+    seed: 0,
+    tid: 0,
+    sid: 0,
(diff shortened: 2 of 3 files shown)


sorendog:
To do list:
-Figure out how to do encounter and lead correctly so it isn't throwing me this error 
`Type '"CutecharmF" | "CutecharmM" | "Synchronize" | "None" | null' is not assignable to type '"CutecharmF" | "CutecharmM" | { Synchronize: Nature; } | null'.
  Type '"Synchronize"' is not assignable to type '"CutecharmF" | "CutecharmM" | { Synchronize: Nature; } | null'.`
-add synch nature selection or set up code to put selected nature with synch when submitting to filter_4static so it is inputted correctly (should fix the synch part of the error i am getting)
-make github happy so i can commit this without the no verify when i git push

Extras:
- Update gen 4 static guides
- make connections so tool is accessible

sorendog:
@zaksabeast i made some notes so maybe will help you with helping me but I am throwing in the flag on this last error. Can you help me figure it out or guide me on what to look up? Thank you

zaksabeast:
@sorendog Great question - you're very close!

Here's some info for how I'd tackle this.  Let me know if this explanation doesn't quite make sense or if you'd like some more info.

## Root problem

It looks like two requirements are conflicting with each other right now:
- Rust generates `LeadAbilities` to be a union of strings and one object
- `FormikSelect` wants values to be a union of strings

If we create a list of strings for `FormikSelect`, but where do we get the nature from for the rust code?  If we use objects for rust, how can we get a list of strings for the select?

We need a solution that:
- Uses a list of strings for `FormikSelect`
- Captures a sync nature from the user
- Combines the two together for rust

## Possible solution

### Creating a list of strings for FormikSelect

First, I'd create a list of sync options for the validator:
```tsx
const leadAbilities = ["None", "CutecharmF", "CutecharmM", "Synchronize"] as const;

const Validator = z
  .object({
    // other fields
    lead: z.enum(leadAbilities),
  })
  .merge(pkmFilterSchema);
```

That will let us make a list of string options to make `FormikSelect` happy:
```tsx
type FormState = z.infer<typeof Validator>;

type LeadAbilityOption = FormState["lead"];

const leadAbilitiesOpts: { label: string; value: LeadAbilityOption }[] = [
  { label: "None", value: "None" },
  { label: "Cute Charm (F)", value: "CutecharmF" },
  { label: "Cute Charm (M)", value: "CutecharmM" },
  {
    label: "Synchronize",
    value: "Synchronize",
  },
];

const fields: Field[] = [
  // other fields
  {
    label: "Lead Ability",
    input: (
      <FormikSelect<FormState, "lead">
        name="lead"
        options={leadAbilitiesOpts}
      />
    ),
  },
];
```

### Capturing the Sync Nature

Next, I'd add a `syncNature` field to capture the nature:
```tsx
const Validator = z
  .object({
    // other fields
    syncNature: z.enum(natures),
  })
  .merge(pkmFilterSchema);

const fields: Field[] = [
  // other fields
    {
    label: "Sync Nature",
    input: (
      <FormikSelect<FormState, "syncNature">
        name="syncNature"
        options={natureOptions.required}
      />
    ),
  },
];
```

We can have this be required because a user who has chosen sync will always need a nature, and a user without sync will have this field ignored either way.

If we made it optional, our validation wouldn't guarantee a sync nature is selected, so we'd have to deal with what happens if the user somehow found a way to select sync without a nature.

### Combining the data for rust

Lastly, I'd add a new function that takes the form state and returns the version formatted for rust:
```tsx
const denormalizeLead = (opts: FormState): RngToolsLeadAbilities => {
  // Convert the lead ability to the format expected by the rngTools
}

const onSubmit = async (opts) => {
  const results = await rngTools.filter_4static({
    // other fields
    lead: denormalizeLead(opts)
  })

  // other submit logic
}
```

## Bonus

As a bonus: try hiding the sync nature field if "Synchronize" isn't the selected option.  As a hint for how we currently handle this elsewhere: `fields` will need to become a `getFields` function.

sorendog:
@zaksabeast 

Thanks so much for the detailed response — it’s helped a ton and I’ve made most of the necessary changes! I did want to approach the Synchronize nature logic a little differently than you described.

My thought was that users might appreciate only selecting a nature once. Since Gen 4 Synchronize works properly and doesn’t affect IVs based on nature, I was hoping to make the nature field conditionally required — only when Synchronize is selected. From what I found, this might be possible using 'superRefine' with Zod.

I couldn’t quite get it working, though. I’m a bit stuck and figured I’d ask:

-Do you think this approach would improve UX?

-If so, is 'superRefine' alone enough to make the nature field conditionally required, or is there more to it?

Any pointers would be greatly appreciated! And sorry this turned into a bit of a long message 😅

sorendog:
I was gonna use this in my own environment to just try it out cause i was excited that this works now- and discovered characteristics is broken. it needs a null/none option as right now you are forced to choose a character which then limits the options. I can try to fix this tomorrow. (later today?? gosh i need to go to sleep) review on everything else will likely be helpful still though because i expect i have a lot of little things i missed ^^

src/types/nature.ts
```diff
@@ -30,4 +29,6 @@ export const nature = [
   "Sassy",
   "Careful",
   "Quirky",
-] as const satisfies Nature[];
+] as const;
+
+export type Nature = (typeof nature)[number];
```
zaksabeast: Let's continue to import `Nature` from `~/rngTools` and use `as const satisfies Nature[]` to bind this list to `rngTools`.

If we want to re-export `Nature`, we can add this to the top of the file:
```ts
export type { Nature } from "~/rngTools";
```

rng_tools/src/generators/gen4/stationary/generator.rs
```diff
@@ -24,7 +24,7 @@ pub struct Gen4StaticOpts {
     pub filter_characteristic: Option<Characteristic>,
     pub game: Option<GameVersion>,
     pub encounter: Static4Species,
-    pub lead: Option<LeadAbilities>,
+    pub lead: LeadAbilities,
```
zaksabeast, line 27: Let's make the same change to `Gen4SearcherOpts` so we don't need to match `None` and `Some(LeadAbilities::None)` in `gen4_method_j` and `gen4_method_k`.

src/rngToolsUi/gen4/gen4types.ts
```diff
@@ -0,0 +1,74 @@
+export const GameVersion = [
+  "Diamond",
+  "Pearl",
+  "Platinum",
+  "HeartGold",
+  "SoulSilver",
+] as const;
+
+export const StaticEncounterId = [
```
zaksabeast: Let's rename this to `StaticEncounterSpecies` since we tend to use `species` elsewhere in the codebase.

src/rngToolsUi/gen4/gen4stationary.tsx
```diff
@@ -0,0 +1,283 @@
...
+  return fields;
+};
+
+export type RustLead =
+  | "None"
+  | "CutecharmF"
+  | "CutecharmM"
+  | { Synchronize: Nature };
+
+export const Filter_4static = () => {
```
zaksabeast: Let's rename this to something like `Static4Generator` or `GenerateStatic4` (at some point I'll go back and make component names have a consistent structure).

Here's my rationale:
- React components typically don't have underscores
- This component always generates static Pokemon, but a user might not always be filtering

Thoughts?

src/rngToolsUi/gen4/gen4stationary.tsx
```diff
@@ -0,0 +1,283 @@
...
+    });
+  }
+  return fields;
+};
+
+export type RustLead =
+  | "None"
+  | "CutecharmF"
+  | "CutecharmM"
+  | { Synchronize: Nature };
```
zaksabeast: Let's import `LeadAbilities` from `~/rngTools` instead of redefining it here.

That way if rust is ever updated with new leads, our TS code will already have the correct types.

If we redefine the same type here, any updates will need to be duplicated here.

src/rngToolsUi/gen4/gen4stationary.tsx
```diff
@@ -0,0 +1,283 @@
...
+  const onSubmit = React.useCallback<RngToolSubmit<FormState>>(async (opts) => {
+    let lead: RustLead;
+    if (opts.lead === "Synchronize") {
+      if (opts.synch_nature === null) {
+        throw new Error("Nature is required for Synchronize.");
+      }
+      lead = { Synchronize: opts.synch_nature };
+    } else {
+      lead = opts.lead;
+    }
```
zaksabeast: Let's pull this logic into a new function that uses `match` with `exhaustive`:
```tsx
const getLeadAbility = (opts: FormState): LeadAbilities => {
  return match<FormState, LeadAbilities>(opts)
    .with({ lead: "Synchronize" }, (matched) => ({
      Synchronize: matched.synch_nature,
    }))
    .with({ lead: "CutecharmF" }, () => "CutecharmF")
    .with({ lead: "CutecharmM" }, () => "CutecharmM")
    .with({ lead: "None" }, () => "None")
    .exhaustive();
};
```

Then below we can use `lead: getLeadAbility(opts)`.

In the future if we update the lead abilities in rust, `match` with `exhaustive` will give a type error if we don't explicitly handle the new lead ability.

This ensures we don't have accidental bugs from assuming we always want to set the `lead` to `opts.lead`.

`getLeadAbility` can also be put in a different file if we need to import it elsewhere in the short term.

src/rngToolsUi/gen4/gen4stationary.tsx
```diff
@@ -0,0 +1,283 @@
...
+  { label: "None", value: "None" },
+  { label: "Cute Charm (F)", value: "CutecharmF" },
+  { label: "Cute Charm (M)", value: "CutecharmM" },
+  {
+    label: "Synchronize",
+    value: "Synchronize",
+  },
+];
+
+const characteristics = [
```
zaksabeast: It looks like we have this list defined in two places.  Let's move it to the new `gen4Types.ts` file you created so it can be shared.

zaksabeast (changes requested):
Looking good so far!  Can't wait to use it now that I've started a new gen 4 game.

src/rngToolsUi/gen4/gen4stationary.tsx
sorendog: I will rename it to Static4Generator - that seems to be more in line with how the rust functions get named so that could be helpful.

src/types/nature.ts
sorendog: I think I understand with this, I am going to remove it and since it doesn't seem to break anything leave it as it was.

sorendog:
All changes requested finished. I even added getleadability to the types file i made so it can be used for other tools and it is in its own place!

src/types/nature.ts
```diff
@@ -30,4 +29,4 @@ export const nature = [
   "Sassy",
   "Careful",
   "Quirky",
-] as const satisfies Nature[];
+] as const;
```
zaksabeast: Let's add `satisfies Nature[];` back - that'll have TS error if we accidentally modify the list to include a typo in one of the natures.

zaksabeast (changes requested):
One super tiny change, then this is ready to merge!

src/types/nature.ts
sorendog: my bad! have this committed back to that in just a sec

zaksabeast (approved):
Thank you!

PR #343, opened 2025-06-20
Gen3 Wild: Add cycle counting to the generator 

RainingChain:


RainingChain:
The next step is creating UI for calling the searcher. See https://docs.google.com/document/d/1quYGRFpSMhZg3EpiWC01sAhY_MfKLIJAvSIlIrzqeX4/edit?tab=t.0#heading=h.928up459747p

rng_tools/src/generators/gen3/wild/generator.rs
```diff
@@ -39,53 +46,247 @@ pub struct Wild3GeneratorOptions {
...
+            mod25: 0,
+        }
+    }
+}
+
+impl CycleCount {
+    pub fn add(&mut self, cycle: usize, reason: &str) {
+        self.cycle += cycle;
+        if VERBOSE {
+            println!("cycle += {} (total: {}) : {}", cycle, self.cycle, reason);
```
zaksabeast: It looks like this is here to either assert some known knowledge or help debug a problem.  Both of those seem like they should be test cases to make sure we don't accidentally modify logic in a way that breaks what this logging is trying to accomplish.

Typically printing in a library isn't good practice - library users don't want unexpected logs.  While we are the user in this case, we don't want to accidentally send these to the user's browser.  We currently don't have any checks to make sure `VERBOSE` isn't set to `true` in the future by mistake.

Is there a reason this needs to be a `println` instead of a test?

If we super need to keep this, let's either put it behind a feature flag in cargo we need to explicitly enable or `#[cfg(debug_assertions)]`.

rng_tools/src/generators/gen3/wild/generator.rs
```diff
@@ -94,18 +295,24 @@ pub fn generate_gen3_wild(
...
                 encounter_slot,
                 pid_low,
                 required_gender,
                 required_nature,
+                CycleRange::from_start_len(cycle, method3_range),
             ) {
                 results.push(gen_mon_wild3);
             }
         }
+        cycle += (
```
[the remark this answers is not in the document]
zaksabeast: Let's ignore this.

rng_tools/src/generators/gen3/wild/generator.rs
RainingChain: @zaksabeast In the current shape of the generator, we can't add a test for that. My end plan is to convert the logs and other debug data in its own optional result data structure. In the meantime, I can just remove the println! and rename `reason` with `_reason`.

zaksabeast: That sounds like a solid end plan - sounds good to me!

zaksabeast (approved):
Awesome, thank you!

PR #350, opened 2025-07-05
Add ColumnGroup support and dynamic columns (getColumns)

RainingChain:
@zaksabeast Only check the commit [ColumnGroup + getColumns](https://github.com/zaksabeast/PokemonRNGGuides/commit/4ba8ce0d2eabd5a818cda097a2eaf1b7d86bdf1f)

image


src/components/resultTable.tsx
```diff
@@ -18,6 +18,13 @@ export type ResultColumn<T> = keyof T extends string
     }[keyof T]
   : never;
 
+export type ResultColumnGroup<T> = {
+  title: string;
+  columns: ResultColumn<T>[];
+};
+
+export type ResultColumnsType<T> = (ResultColumn<T> | ResultColumnGroup<T>)[];
```
zaksabeast: It seems a bit redundant to append `Type` to types.

Could this be called `ResultColumns`?

We could also rename `ResultColumn` to `SingleResultColumn` and export `ResultColumn` as the new union:
```tsx
type SingleResultColumn<T> = keyof T extends string
  ? {
      [K in keyof T]: {
        title: string;
        dataIndex: K;
        monospace?: boolean;
      } & (T[K] extends string | number | undefined
        ? { render?: (value: T[K], values: T) => React.ReactNode }
        : { render: (value: T[K], values: T) => React.ReactNode });
    }[keyof T]
  : never;

export type ResultColumnGroup<T> = {
  title: string;
  columns: SingleResultColumn<T>[];
};

export type ResultColumn<T> = (SingleResultColumn<T> | ResultColumnGroup<T>);
```

src/components/resultTable.tsx
```diff
@@ -18,6 +18,13 @@ export type ResultColumn<T> = keyof T extends string
     }[keyof T]
   : never;
 
+export type ResultColumnGroup<T> = {
+  title: string;
+  columns: ResultColumn<T>[];
+};
```
zaksabeast: Let's add a `type` property to each: an optional `type?: "single"` to the single column and a required `type: "group"` to the group column:
```tsx
export type ResultColumn<T> = keyof T extends string
  ? {
      [K in keyof T]: {
        type?: "single";
        title: string;
        dataIndex: K;
        monospace?: boolean;
      } & (T[K] extends string | number | undefined
        ? { render?: (value: T[K], values: T) => React.ReactNode }
        : { render: (value: T[K], values: T) => React.ReactNode });
    }[keyof T]
  : never;

export type ResultColumnGroup<T> = {
  type: "group";
  title: string;
  columns: ResultColumn<T>[];
};
```

Tagged unions like this safer to work with.

Without tags, TS can get confused about overlapping types.  In the existing code, I can add `monospace: true,` to group columns, even though that shouldn't be possible.

Checks like `"columns" in column` are quite brittle and should only be used as a last resort. If we rename `columns` to something else, our logic will be wrong but we won't get type errors telling us where to update our code.

Other problems can come up if we have overlapping unions in the future.

Adding `type` with a constant prevents these issues.

src/components/resultTable.tsx
```diff
@@ -39,23 +46,50 @@ const applyMonospace = <Record extends tst.O.Object>(
...
+        column.columns = column.columns.map(applyMonospace);
+        return column;
+      }
+      return applyMonospace(column);
+    });
+  }, [props.columns]);
+
+  const children = React.useMemo(() => {
+    return columns.map((column) => {
+      if ("columns" in column) {
```
zaksabeast: Let's update these two `"columns" in column` checks to be `column.type === "group"`.

src/components/rngToolForm.tsx
```diff
@@ -89,6 +97,10 @@ export const RngToolForm = <
           return <FormFieldTable fields={fieldsToUse} />;
         })();
 
+        const columnsToUse = (() => {
+          return columns ?? getColumns?.(formik.values) ?? null;
+        })();
```
zaksabeast: Is there a reason to not do this:
```tsx
const columnsToUse = columns ?? getColumns?.(formik.values) ?? null;
```

src/components/resultTable.tsx
RainingChain: @zaksabeast I followed the same naming convention as `antd`. 

See `.\node_modules\antd\es\table\interface.d.ts` 
`export type ColumnsType<RecordType = AnyObject> = (ColumnGroupType<RecordType> | ColumnType<RecordType>)[];`

I agree that naming convention isn't ideal.

RainingChain:
@zaksabeast Note: Wait until the Gen3 Wild searcher is merged before merging this one.

zaksabeast (approved):
Thank you!

RainingChain:
@zaksabeast It's ready to be merged.

zaksabeast (approved):
Thank you!

PR #351, opened 2025-07-07
HGSS seed time stuff

sorendog:
I was hoping I would get to finish this all tonight and not do a draft pull request but it is late now.

Adds:
Roamer location logic for HGSS,
Elm call logic,
functions to call to get hgss seed dates with the elm and roamer data wrapped in instead of coin flips.

as of right now, I just need to double check get_route_k  and elm calls as they cause my test to fail.

### DIFF
--- a/rng_tools/src/generators/gen4/seed_time.rs
+++ b/rng_tools/src/generators/gen4/seed_time.rs
@@ -397,7 +397,17 @@ pub fn hgss_calculate_seedtime(opts: HgssSeedTime4Options) -> Vec<HgssSeedTime4>
 }
 
 fn hgss_calculate_single_month_seedtime(opts: HgssSeedTime4Options) -> Vec<HgssSeedTime4> {
-    let elm = get_elm_calls(opts.seed);
+    let seed = opts.seed;
+    let mut elm_calls_res: Option<Vec<ElmCall>> = None;
+
+    let mut lazy_elm_calls = || match elm_calls_res {
+        Some(ref calls) => calls.clone(),
+        None => {
+            let new_calls = get_elm_calls(seed);
+            elm_calls_res = Some(new_calls.clone());
+            new_calls
+        }
+    };
     let roamer = roamer_check(opts.seed, opts.roamer);
     calculate_seedtime!(from_optional_month:opts)
         .into_iter()
@@ -405,7 +415,7 @@ fn hgss_calculate_single_month_seedtime(opts: HgssSeedTime4Options) -> Vec<HgssS
             convert_seedtime!( s, HgssSeedTime4, {
 
                 roamer: roamer.clone(),
-                elm: elm.clone(),
+                elm: lazy_elm_calls(),
             })
         })
         .collect()
@@ -454,20 +464,20 @@ pub fn dppt_calibrate_seedtime(
     results
 }
 
-fn get_route_j(seed: u32, roamer: &mut RoamerLocation) {
-    let roamer_rand = (seed >> 16) as u16 & 15;
+fn get_route_j(rand: u16) -> u16 {
+    let roamer_rand = rand & 15;
 
-    roamer.location = if roamer_rand < 11 {
+    if roamer_rand < 11 {
         roamer_rand + 29
     } else {
         roamer_rand + 31
-    };
+    }
 }
 
-fn get_route_k(seed: u32, roamer: &mut RoamerLocation) {
-    let roamer_rand = (seed >> 16) as u16 % 25;
+fn get_route_k(rand: u16) -> u16 {
+    let roamer_rand = rand % 25;
 
-    let location = if roamer_rand > 21 {
+    if roamer_rand > 21 {
         if roamer_rand == 22 {
             24
         } else if roamer_rand == 23 {
@@ -477,9 +487,7 @@ fn get_route_k(seed: u32, roamer: &mut RoamerLocation) {
         }
     } else {
         roamer_rand + 1
-    };
-
-    roamer.location = location;
+    }
 }
 
 fn roamer_check(seed: u32, roamer_opts: RoamerSet) -> Vec<RoamerLocation> {
@@ -488,22 +496,22 @@ fn roamer_check(seed: u32, roamer_opts: RoamerSet) -> Vec<RoamerLocation> {
 
     if roamer_opts.raikou {
         let mut roamer = RoamerLocation::new(Species::Raikou);
-        get_route_j(rng.rand::<u32>(), &mut roamer);
+        roamer.location = get_route_j(rng.rand::<u16>());
         results.push(roamer)
     }
     if roamer_opts.entei {
         let mut roamer = RoamerLocation::new(Species::Entei);
-        get_route_j(rng.rand::<u32>(), &mut roamer);
+        roamer.location = get_route_j(rng.rand::<u16>());
         results.push(roamer)
     }
     if roamer_opts.latios {
         let mut roamer = RoamerLocation::new(Species::Latios);
-        get_route_k(rng.rand::<u32>(), &mut roamer);
+        roamer.location = get_route_k(rng.rand::<u16>());
         results.push(roamer)
     }
     if roamer_opts.latias {
         let mut roamer = RoamerLocation::new(Species::Latias);
-        get_route_k(rng.rand::<u32>(), &mut roamer);
+        roamer.location = get_route_k(rng.rand::<u16>());
         results.push(roamer)
     }
     results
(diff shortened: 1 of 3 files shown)


rng_tools/src/generators/gen4/seed_time.rs
```diff
@@ -276,6 +278,222 @@ pub fn dppt_calibrate_seedtime(
...
+
+fn get_route_j(seed: u32, roamer: &mut RoamerLocation) {
+    let roamer_rand = (seed >> 16) as u16 & 15;
+
+    roamer.location = if roamer_rand < 11 {
+        roamer_rand + 29
+    } else {
+        roamer_rand + 31
+    };
+}
```
zaksabeast: Let's return a location instead of mutating the argument.  It might not make much difference in this function, but in general that's preferable when possible.  This makes functions more generic, which makes them easier to reuse and write tests for.

Let's also take the random number instead of the rng state.  That lets us leverage the `rand` method:
```rs
fn get_route_j(rand: u16) -> u16 {
    let roamer_rand = rand & 15;

    if roamer_rand < 11 {
        roamer_rand + 29
    } else {
        roamer_rand + 31
    }
}
```

Down below we can call the function like this:
```rs
roamer.location = get_route_j(rng.rand::<u16>());
```

Or we could initialize the location without any mutations:
```rs
let roamer = RoamerLocation {
  roamer: Species::Raikou,
  location: get_route_j(rng.rand::<u16>())
};
```

rng_tools/src/generators/gen4/seed_time.rs
```diff
@@ -276,6 +278,222 @@ pub fn dppt_calibrate_seedtime(
...
+            26
+        } else {
+            28
+        }
+    } else {
+        roamer_rand + 1
+    };
+
+    roamer.location = location;
+}
```
zaksabeast: Same as above, let's accept an `rng.rand::<u16>()` as the first argument and return the location.

rng_tools/src/generators/gen4/seed_time.rs
```diff
@@ -276,6 +278,222 @@ pub fn dppt_calibrate_seedtime(
...
+        false => efgh.wrapping_add(2000).wrapping_sub(year),
+    };
+
+    if let Some(delay_range) = opts.delay_range {
+        if !delay_range.contains(&delay) {
+            return vec![];
+        }
+    }
+    let elm = get_elm_calls(opts.seed);
+    let roamer = roamer_check(opts.seed, roamer);
```
zaksabeast: Let's use the same approach as `lazy_coin_flip` in the other function.

From what I remember, not every seed has a date time and calculating coin flips for every seed had a noticeable performance cost.  `lazy_coin_flip` ensures we only calculate coin flips for seeds with date times.

It's not as clean, but I recall it being much more performant in some cases.

rng_tools/src/generators/gen4/seed_time.rs
```diff
@@ -276,6 +278,222 @@ pub fn dppt_calibrate_seedtime(
...
+        year: opts.year,
+        month: None,
+        second_range: opts.second_range,
+        delay_range: Some(opts.delay_range),
+        find_first: true,
+    };
+    hgss_calculate_seedtime(opts, roamer).into_iter().next()
+}
+
+pub fn hgss_calculate_seedtime(opts: SeedTime4Options, roamer: RoamerSet) -> Vec<HgssSeedTime4> {
```
zaksabeast: At some point I'd like to fully replace `rngTools` in TypeScript with the new `multiWorkerRngTools`.

To make `multiWorkerRngTools` generic, it only works with functions that take a single parameter.

Since we need an HGSS specific seed time type, let's make `HgssSeedTime4Options` as well and include roamers.

rng_tools/src/generators/gen4/seed_time.rs
```diff
@@ -276,6 +278,222 @@ pub fn dppt_calibrate_seedtime(
...
+                    find_first: opts.find_first,
+                },
+                roamer,
+            )
+        })
+        .take(limit)
+        .collect()
+}
+
+fn hgss_calculate_single_month_seedtime(
```
zaksabeast: Not a required change, but here's a challenge in case you're interested (totally fine if not):

## An Observation

There's quite a bit of duplicate code between hgss and dppt.  It looks like 95% is copied and only 5% is unique.

Normally we wouldn't want as much duplicate code because:
- Bugs are easier to introduce (change one place, but forget another)
- Harder to make changes (need to changes lots of places, not just one)
- Code is harder to read (not easy to spot subtle differences)

Among other reasons.

However code for older Pokemon games has a little more leeway:
- The game logic will never change, so our code won't change often.  Even if we refactor, the logic should still pass the same tests.  This is different from other environments that have constantly changing requirements (e.g. a new app feature) - our requirements rarely change.
- The games have _just enough_ differences scattered around that sharing logic tends to hurt readability in a lot of languages.  Typically code (e.g. for some company's app) is planned out so it's easy to reuse and share over time, but a lot of old Pokemon logic was never intended to be shared between games.  Trying to retrofit old code in a way it wasn't intended creates a few issues.

Compared to other coding environments, trying to share Pokemon logic often comes with more downsides and fewer upsides, so more duplicate code than normal is alright.

However rust has a pretty nice way to reduce duplicate code without all the trouble other languages run into.

## Rust Macros

A rust macro generates rust code.  You can think of it like a code template with variables that can be replaced (the macro parameters).

Here's an example:
```rs
macro_rules! say_hello {
    () => {
        println!("Hello!");
        println!("Hello!");
        println!("Hello!");
    };
}
```

If you use the `say_hello!` macro, it's the same as copy/pasting the contents of the macro into the same place you used `say_hello!`.

## Macro parameters

Since macros are templates, we can give it stuff to put in the template:
```rs
macro_rules! say_hello {
    ($name:expr) => {
        println!("Hello, {}!", $name);
        println!("Hello again, {}!", $name);
    };
}
```

Using `say_hello!("Zak")` is the same as copy/pasting this code into your logic:
```rs
println!("Hello, {}!", "Zak");
println!("Hello again, {}!", "Zak");
```

## Looser types with macros

Because it's generating code, it's kind of like you can bend type rules a little bit.

Here's an example `add!` macro that works with any type:
```rs
macro_rules! add {
    ($a:expr, $b:expr) => {{
        $a + $b
    }};
}
```

So we can use it like this:
```rs
add!(1_u32, 2_u32);
add!(10_i64, -5_i64);
```

An `add` _function_ needs to define the types it accepts.  It could only use `u32` _or_ `i64`.  Since macros are code templates, the `add!` macro is just copy pasting the `num + num` code.

You'll still get compiler errors if the generated code isn't type safe, so we're not ignoring type rules, but it lets us write code in a way that feels like types are loosened.

## Types as parameters

Functions can also only accept values, but macros can accept anything, even types:
```rs
macro_rules! print_default {
    ($t:ty) => {{
        let val: $t = Default::default();
        println!("{:?}", val);
    }};
}

fn main() {
    print_default!(u32);
    print_default!(bool);
    print_default!(String);
}
```
The compiled code will be:
```rs
fn main() {
    let val: u32 = Default::default();
    println!("{:?}", val);
    let val: bool = Default::default();
    println!("{:?}", val);
    let val: String = Default::default();
    println!("{:?}", val);
}
```

Macros can pretty much accept anything: types, function names, logic, primitive values, etc.  Functions can't do any of that.  That makes macros super powerful.

## Let's tie this together

Since macros generate code and can accept function names, logic, and types as arguments, we can write a macro that generates entire functions, or partial functions.

Imagine a macro that generates the `hgss_calculate_single_month_seedtime` function, but:
- The name comes from a parameter (so we can prefix it with `hgss` or `dppt`)
- The types comes from parameters (for the game specific input and output types)
- The tiny bit of logic for elm/roamer/coin flips can be a parameter

That way the 95% shared logic is in the macro, and the 5% of differences between games come from using the macro for each set of games.

This helps reduce code while avoiding some of the problems we'd normally run into with old Pokemon logic.

## Up for the challenge?

No worries if not - I'll merge this PR either way.  But I wanted to share this as an option in case you're interested.

zaksabeast (changes requested):
Took an initial pass at this PR.  Looks good!  Just a few small requests.

rng_tools/src/generators/gen4/seed_time.rs
sorendog: That makes sense - Do all functions need to be only single parameter or will things like my roamer_check be okay since it is kind of off of the main seed finder function pathway?

rng_tools/src/generators/gen4/seed_time.rs
sorendog: This was a fun read and I enjoyed looking into macro's for rust - i did decide to take this challenge. I made a few and added them to functions, i in general did restructing and organizing to help me better able to process what to do. Does it look like I did as you thought or would you like some more of the duplicate functions to be macro'd?

rng_tools/src/generators/gen4/seed_time.rs
```diff
@@ -225,28 +230,210 @@ macro_rules! coin_flips {
...
-    pub datetime: RngDateTime,
-    pub delay: u32,
-    pub coin_flips: Vec<CoinFlip>,
+#[macro_export]
+macro_rules! calculate_seedtime {
+    (from_optional_month: $opts:expr) => {{
+        let opts = $opts;
+        let month = match opts.month {
+            Some(m) => m,
+            None => 1, // or panic!("Missing month in opts")
```
zaksabeast: You're right to not panic here 👍 

We typically don't want to panic in libraries, and we should treat `rng_tools` like an external library. Anyone who contributes to `rng_tools` might not be the same person who builds a UI with it.

Libraries shouldn't have unexpected surprises. Imagine building a mobile app with an http library that panics - users would randomly have their app crash/close. That wouldn't be a good experience, and the app devs couldn't prevent it since the panic is in the library, not their own code. Unless they read the entire source, they won't know the library can panic in the first place.

When libraries run into unexpected situations they should either return an error to let the caller decide how to handle the problem, or handle it with reasonable default behavior. In the first case, the app devs could show an error to the user for example. Default behavior should only be used when that's what a caller would expect - it should be intuitive.

Although in this case, I suspect something is missing. When calculating a seedtime, we either want a specific month or we want to iterate over all months. There shouldn't be a time we try to calculate a seedtime without a month.

rng_tools/src/generators/gen4/seed_time.rs
zaksabeast: Wow, great job! Yeah, this is pretty much what I was expecting. It's easy to follow and removes the duplication.

One suggestion is to move the `convert_seedtime` into the `calculate_seedtime` macro. `calculate_seedtime` creates a `Vec<SeedTime4>`, then we map over it and create a second vector.

If `calculate_seedtime` uses `convert_seedtime` like this:
```rs
let seedtime = SeedTime4 {
    seed: opts.seed,
    delay,
    datetime,
};
let result = convert_seedtime!(
    seedtime,
    $target,
    {
        $($extra_field: $value),*
    }
);
results.push(result);
```

Then we can use it and only return the result vector:
```rs
fn dppt_calculate_single_month_seedtime(opts: SeedTime4SingleMonthOptions) -> Vec<DpptSeedTime4> {
    let seed = opts.seed;
    let month = opts.month;
    let mut coin_flips_res: Option<Vec<CoinFlip>> = None;
    let mut lazy_coin_flip = || match coin_flips_res {
        Some(ref flips) => flips.clone(),
        None => {
            let new_flips = coin_flips(seed);
            coin_flips_res = Some(new_flips.clone());
            new_flips
        }
    };
    calculate_seedtime!(core:opts, month, DpptSeedTime4, {
        coin_flips: lazy_coin_flip(),
    })
}
```

Thoughts?

rng_tools/src/generators/gen4/seed_time.rs
zaksabeast: Only `#[wasm_bindgen]` functions that will be used in the web app need a single parameter. Everything else is fine as is.

zaksabeast (changes requested):

rng_tools/src/generators/gen4/seed_time.rs
sorendog: That makes sense to me - tbh I'd probably uninstall an app if it crashed too often or froze up. It makes it seem like the app isn't coded well when maybe it is. It just had a panic in a place they thought it wasn't possible to reach, but in actuality, it was. 

Anyway - with regards to this, I'll be honest it exists because I didn't want to make another hgss exclusive struct. The hgss function path way is missing the first function dppt has which converts from one struct to another basically - because of that the first function for hgss uses a struct that has option(u32) for month as its where a user could enter the month(s) they want. The function- if I was understanding correctly- modifies it so when we go to search single month, only 1 month is being passed along. But I'm using the struct that uses an option still- so the macro would error if I didnt have match arms for all possibilities. It shouldn't be possible to send without a month if im understanding the functions all correctly. 

I hope this explanation helps.

zaksabeast: Yeah you're right - that's a good reason to use `Option` here.

Instead of defaulting to `1`, perhaps we should have early return an empty vec.

If in 6 months us or someone new to the project accidentally adds a bug where `None` is possible, we might not catch the bug if we default to `1`. If we always return 0 results for an invalid input, we have a higher chance of noticing a problem.

You might know this already, but using `return` in a match branch will return the entire function, not just the match:
```rs
fn add_1_except_for_5(num: u32) -> u32 {
    let result = match num {
        5 => return 5,
        num => num + 1,
    };

    println!("Result is {}", result);

    result
}

fn main() {
    add_1_except_for_5(2);
    add_1_except_for_5(5);
    add_1_except_for_5(6);
}
```
This logic prints `Result is 3` and `Result is 7`, but no log for `5` since it has an early return in the match.

Removing the return like this:
```rs
let result = match num {
    5 => 5,
    num => num + 1,
};
```
Now prints `Result is 5`.

For this PR, we can do this to return an empty list early:
```rs
let month = match opts.month {
    Some(month) => month,
    None => return vec![],
};
```

Thoughts on this approach?

sorendog: I think this is a good approach - it will make it easier to tell if there is an issue with something new later if it is doing a empty vec instead. I will get this update to do that instead!

sorendog:
I think this is ready for another review when your able to!

zaksabeast (approved):
Thank you!

PR #352, opened 2025-07-07
Add Hidden Power, Gen3 Max Size and Gen3 PID Speed filters

RainingChain:
Max size filter is only displayed if the size of the Pokémon matters in RSE (Lotad, Seedot, Barboach, Shroomish).

Possible improvements:
 - Fix the `//@ts-expect-error TODO` related with formik typing validation
 - Change the type of `PkmFilter.hidden_power` to be optimized for performance. This means that the Rust and Typescript data structure must be different.

image
image


src/components/pkmFilter.tsx
```diff
@@ -41,6 +43,15 @@ export type PkmFilterFields = {
   [Key in keyof PkmFilter as `filter_${Key}`]: undefined extends PkmFilter[Key]
     ? tst.U.Exclude<PkmFilter[Key], undefined> | null
     : PkmFilter[Key];
+} & {
+  filter_hidden_power: tst.U.Exclude<PkmFilter['hidden_power'], undefined>,
+};
+
+//NO_PROD bad
+export type PkmFilterNull = {
```
RainingChain: @zaksabeast Here

rng_tools/src/pkm/hidden_power.rs
```diff
@@ -0,0 +1,63 @@
+use super::Ivs;
+use super::pokemon_type::PokemonType;
+use serde::{Deserialize, Serialize};
+use tsify_next::Tsify;
+
+#[derive(Debug, Default, Clone, PartialEq, Tsify, Serialize, Deserialize)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub struct HiddenPowerFilter {
+    pub active: bool,
+    // TODO: for performance, [bool, POKEMON_TYPE_COUNT] would be better
```
zaksabeast, line 10: Is the concern memory allocations? As long as we're not copying the filter everywhere, this shouldn't be much of an issue.

If there are other concerns, a bitfield with named bits that implements `Iterator` might be a more readable, performant, and use strict types in the interface.

src/components/gen3PkmFilter.tsx
```diff
@@ -0,0 +1,122 @@
...
+    min_cycle_count: 18,
+    max_cycle_count: 900,
+  },
+});
+
+const optOut = <T,>(condition: boolean | undefined, value: T): T | null => {
+  return condition === false ? null : value;
+};
+
+const optIn = <T,>(condition: boolean | undefined, value: T): T | null => {
```
zaksabeast: Let's turn these into utilities and share them between the generic PkmFilter and this file.

src/components/hiddenPowerInput.tsx
```diff
@@ -0,0 +1,86 @@
...
+import React from "react";
+import { PkmFilterFields } from "./pkmFilter";
+import { Paths } from "~/types";
+
+type UiHiddenPowerFilter = {
+  active: boolean;
+  pokemon_types: PokemonType[];
+  min_bp: number;
+  max_bp: number;
+};
```
zaksabeast: Let's derive this from the schema so any updates only need to happen in one place:
```ts
type UiHiddenPowerFilter = z.infer<typeof HiddenPowerSchema>;
```

src/components/hiddenPowerInput.tsx
```diff
@@ -0,0 +1,86 @@
...
+}: Props<FormState>) => {
+  const [{ value: active }] = useField<HiddenPowerFilter["active"]>(
+    `${name}.active`,
+  );
+  const fields = React.useMemo(
+    () => [
+      {
+        label: "Type",
+        input: (
+          <FormikSelect<DummyType, "pokemon_types">
```
zaksabeast: Should be fixed with my latest push:
```ts
<FormikSelect<PkmFilterFields, "filter_hidden_power.pokemon_types">
```

zaksabeast (changes requested):
Just a few tiny tweaks, otherwise it looks great!

rng_tools/src/pkm/hidden_power.rs
RainingChain: The goal is to avoid calling contains() which is O(n). However, changing that for a bitfield requires many changes on the client side. I'll keep it for another PR.

PR #354, opened 2025-07-09
Refactoring: Use Default::default() to reduce boilerplate

RainingChain:
This will simplify the next PR that will add Hidden Power and Max Size filter.


zaksabeast (approved):
Feels so much better! Thank you!

PR #355, opened 2025-07-10
Lead PID Speed Calculator

RainingChain:
image

### DIFF
--- a/src/rngToolsUi/gen3/pidSpeedCalculator.tsx
+++ b/src/rngToolsUi/gen3/pidSpeedCalculator.tsx
@@ -1,8 +1,9 @@
 import React from "react";
-import { Input, Typography, Flex, FormFieldTable } from "~/components";
+import { Typography, Flex, FormFieldTable } from "~/components";
 import { rngTools } from "~/rngTools";
 import { match, P } from "ts-pattern";
 import { formatProbability } from "~/utils/formatProbability";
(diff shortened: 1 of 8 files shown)


src/rngToolsUi/gen3/pidSpeedCalculator.tsx
```diff
@@ -0,0 +1,91 @@
...
+        return `${category}. ${ranking < 0.5 ? fasterThan : slowerThan}`;
+      });
+  }, [speed, ranking]);
+
+  const fields = React.useMemo(() => {
+    return [
+      {
+        label: "Lead PID",
+        input: (
+          <Input
```
zaksabeast: Let's use a `NumberInput` to make sure this follows the same styles and rules as any other number inputs:
```tsx
<NumberInput
  numType="hex"
  name="Seed"
  placeholder="1234abcd"
  onChange={async (pid) => {
    if (pid == null) {
      setSpeed(null);
      setRanking(null);
      return;
    }

    const pidSpeed = await rngTools.calculate_pid_speed(pid);
    const ranking =
      await rngTools.calculate_pid_speed_ranking(pidSpeed);
    setSpeed(pidSpeed);
    setRanking(ranking);
  }}
/>
```

I just pushed a few changes to export `NumberInput` from `~/components` and optionally track an internal state when `value` isn't provided.

zaksabeast (approved):
Thank you!

PR #357, opened 2025-07-10
Gen3 Wild: Support all encounter types

RainingChain:


rng_tools/src/generators/gen3/wild/generator_cycle_tests.rs
```diff
@@ -35,28 +37,14 @@ mod test {
...
                 pid: 1459093362,
-                ivs: Ivs {
-                    hp: 13,
-                    atk: 14,
-                    def: 2,
-                    spa: 20,
-                    spd: 14,
-                    spe: 15,
-                },
+                ivs: Ivs::new(13, 14, 2, 20, 14, 15),
```
zaksabeast, line 40: Not a requirement, but in case it's helpful we also have an `ivs!` macro for when the ivs are known ahead of time like tests.

PR #360, opened 2025-07-29
Add optional key prop on Column and Field, and support ReactNode title and label

RainingChain:
```
const test1:Field = {
   label:<Something />,
   key:"something",
};  // no error

const test2:Field = {
   label:"something",
};  // no error

const test3:Field = {
   label:<Something />,
};  // error. key must be provided
```

### DIFF
--- a/src/components/formFieldTable.tsx
+++ b/src/components/formFieldTable.tsx
@@ -23,7 +23,7 @@ export type Field = {
     }
   | {
       label: string;
-      key?: string; // title will be used as the key
+      key?: undefined; // title will be used as the key
     }
 );
 
@@ -53,7 +53,7 @@ export const FormFieldTable = ({ fields }: Props) => {
                 </Tooltip>
               );
 
-            const keyToUse = typeof key === "string" ? key : (label as string);
+            const keyToUse = key == null ? label : key;
             if (direction === "row") {
               return (
                 <tr key={keyToUse}>
--- a/src/components/formFieldTable.tsx
+++ b/src/components/formFieldTable.tsx
@@ -12,12 +12,20 @@ const LabelTd = styled.td({
 });
 
 export type Field = {
-  label: string;
   hide?: boolean;
   tooltip?: string;
   input: React.ReactNode;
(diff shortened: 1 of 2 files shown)


src/components/formFieldTable.tsx
```diff
@@ -27,43 +35,47 @@ export const FormFieldTable = ({ fields }: Props) => {
...
+              ) : (
+                <Tooltip title={tooltip}>
+                  <Flex gap={8}>
+                    <div>{label}</div>
+                    <Icon name="InformationCircle" size={16} />
+                  </Flex>
+                </Tooltip>
+              );
+
+            const keyToUse = typeof key === "string" ? key : (label as string);
```
zaksabeast: I _think_ the reason TS can't tell `label` should be a string is because `React.ReactNode` is a union that includes a string.

That means a type with this shape:
```ts
{
    label: string;
    key?: string;
}
```
can also satisfy:
```ts
{
  label: React.ReactNode;
  key: string;
}
```
A non-string `key` _should_ make `label` a string, but the overlap of the union seems to be confusing TS.

Instead we can use
```ts
  | {
      label: React.ReactNode;
      key: string;
    }
  | {
      label: string;
      key?: undefined; // title will be used as the key
    }
```

Since `key` is unique per variant, we have a discriminator we can safely use without a cast:
```ts
const keyToUse = key == null ? label : key;
```

src/components/resultTable.tsx
```diff
@@ -71,20 +88,28 @@ export const ResultTable = <Record extends tst.O.Object>(
...
         );
       }
-      return (
-        <Table.Column {...column} key={column.dataIndex + " " + column.title} />
-      );
+
+      const colKey =
+        typeof column.key === "string"
+          ? column.key
+          : column.dataIndex + " " + (column.title as string);
```
zaksabeast: Similar to the other file, let's drop the `as string` here by updating the union to use `key?: undefined`.

zaksabeast (approved):
Thanks!

PR #361, opened 2025-07-29
Wild3: Provide game data as input to generator and searcher

RainingChain:
This is required for : to support multiple species


zaksabeast (approved):
Looks good!

PR #362, opened 2025-07-30
Wild3: Tool to generate all methods at once

RainingChain:
image

### DIFF
--- a/guides/Gen 3/Emerald/Gen 3 Wild.mdx
+++ b/guides/Gen 3/Emerald/Gen 3 Wild.mdx
@@ -7,20 +7,30 @@
   isRoughDraft: true
 ---
 
+## Searcher
+
+**Use case**: Determine the best setup (map, advance, lead) to obtain the wanted target Pokémon.
+
 <Wild3SearcherFindTarget game="emerald" />
 
-<Gen3PidSpeedCalculator />
+<br />
+
+## Generate All Methods At Once
+
+**What it does**: Given a setup (map, advance, lead), calculate the Pokémon outcome for every method (Wild1-Wild5) and their likelihood.
+
+**Use case**: Determine if the likehood of hitting your target Pokémon is good enough with your actual lead.
+
+<Wild3MethodDistribution game="emerald" />
 
 ## Coming soon
 
 - Textual guide for the Emerald Wild Searcher.
 - Calibration tool.
-- Add a filter for the lead cycle speed to help obtain a lead with the desired cycle speed.
-- Add a filter for Hidden Power.
 - Performance improvements.
 
 ## Credits
 
-- RainingChain: Research about cycle range to estimate method likelihood. Creation of the multi-lead searcher, and the web UI tool.
+- RainingChain: Research about cycle range to estimate method likelihood. Creation of the multi-lead searcher, multi-method generator, and the web UI tool.
 - Sorendog: Original Wild generation tool.
 - Shao: Function to calculate cycles taken by modulo operations.
--- a/src/rngToolsUi/gen3/wild/wild3MethodDistribution.tsx
+++ b/src/rngToolsUi/gen3/wild/wild3MethodDistribution.tsx
@@ -2,7 +2,6 @@ import {
   rngTools,
   Species,
   Wild3SearcherResultMon,
-  Wild3SearcherCycleData,
   Gen3Lead,
   Gen3Method,
   Wild3MethodDistributionResult,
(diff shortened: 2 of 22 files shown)


guides/Gen 3/Emerald/Gen 3 Wild.mdx
```diff
@@ -7,20 +7,30 @@
   isRoughDraft: true
 ---
 
+## Searcher
+
+**Use case**: Determine the best setup (map, advance, lead) to obtain the wanted target Pokémon.
```
RainingChain, line 12: Note that there are some bugs with the generator logic that cause the results to incorrect in some cases. I will fix them in another PR.

RainingChain:
Note that there are some bugs with the generator logic that cause the results to incorrect in some cases. I will fix them in another PR.

src/rngToolsUi/gen3/wild/wild3MethodDistribution.tsx
```diff
@@ -0,0 +1,449 @@
...
+  if (leadType !== "Egg") {
+    fields.push({
+      label: "Lead Speed",
+      input: (
+        <FormikRadio<FormState>
+          name="leadSpeedType"
+          options={leadSpeedTypes.slice(0)}
+        />
+      ),
+    });
```
zaksabeast, line 170: I added a `hide` property to `Field` - what are thoughts on using that here?

This would keep a flat list instead of nested ifs, which might be easier to read:

```tsx
const getFields = (
  map_id: string,
  leadType: Gen3Lead,
  leadSpeedType: LeadSpeedType,
  leadCycleSpeed: number,
): Field[] => {
  const encounterTypes = emeraldWildGameData.encounter_tables
    .filter((table) => table.map_id === map_id)
    .map((table) => table.encounter_type);

  return [
    {
      label: "Map",
      input: (
        <FormikSelect<FormState, "map">
          name="map"
          options={toOptions(emeraldWildGameData.maps, formatMapName)}
        />
      ),
    },
    {
      label: "Encounter Type",
      input: (
        <FormikSelect<FormState, "encounterType">
          name="encounterType"
          options={toOptions(encounterTypes, formatEncounterTypeName)}
        />
      ),
    },
    {
      label: "TID",
      input: <FormikNumberInput<FormState> name="tid" numType="decimal" />,
    },
    {
      label: "SID",
      input: <FormikNumberInput<FormState> name="sid" numType="decimal" />,
    },
    {
      label: "Lead",
      input: (
        <FormikSelect<FormState, "leadTypeIdx">
          name="leadTypeIdx"
          // Limitation: value must be a primitive, so we use the index instead of Gen3Lead.
          options={leadTypeOptions.map((el, idx) => {
            return { label: el.label, value: idx };
          })}
        />
      ),
    },
    {
      label: "Lead Speed",
      hide: leadType === "Egg",
      input: (
        <FormikRadio<FormState>
          name="leadSpeedType"
          options={leadSpeedTypes.slice(0)}
        />
      ),
    },
    {
      label: "",
      key: "From PID",
      hide: leadSpeedType !== "From PID" || leadType === "Egg",
      input: (
        <FormFieldTable
          fields={[
            {
              label: "Lead PID:",
              input: (
                <FormikNumberInput<FormState> name="leadPID" numType="hex" />
              ),
            },
          ]}
        />
      ),
    },
    {
      label: "",
      key: "Custom",
      hide: leadSpeedType !== "Custom" || leadType === "Egg",
      input: (
        <FormFieldTable
          fields={[
            {
              label: "PID modulo cycle count:",
              input: (
                <FormikNumberInput<FormState>
                  name="leadCycleSpeed"
                  numType="decimal"
                />
              ),
            },
          ]}
        />
      ),
    },
    {
      label: "",
      hide: leadType === "Egg",
      input: (
        <FormFieldTable
          fields={[
            {
              label: "Cycle Count:",
              input: `${leadCycleSpeed} cycles`,
            },
          ]}
        />
      ),
    },
    {
      label: "Advance",
      input: <FormikNumberInput<FormState> name="advance" numType="decimal" />,
    },
  ];
};
```

RainingChain: @zaksabeast Tbh, I don't really like the proposed change. If the goal is to flatten the structure, I would prefer either:

1) `...(leadSpeedType !== "From PID" || leadType === "Egg" ? [] : [<FormikX>]`

2) Permit null values in Fields 
`leadSpeedType !== "From PID" || leadType === "Egg" ? null : <FormikX>`
That's what React uses normally.

3) Idem as 2), but use `.flatMap(f => f ? [f] : []);` to remove null before giving the array to the core.


If you really want to add an additional properties, I'd suggest a positive term (ex: display:true) to avoid double negation (hide:false).

zaksabeast: I'm up for the first suggestion.

`FormFieldTable` can already accept `null` for the input (it's a `ReactNode`), but it doesn't hide the label. Devs would have to know a null input means the label would also not appear. While you're right that `null` is a typical way to not render something, is isn't typical/expected for surrounding elements to also disappear.

One benefit of a prop is not needing several ternary spreads – it's just a personal preference, but they don't read quite as clean.

I went with `hide` over `display` because we'd need to opt out of an action `display: false` instead of opt into `hide: true`, but I'm not opposed to `display` either.

For now let's go with spreading the single item.

I'm going to merge the PR as is - we can move to spreading at some point.

PR #370, opened 2025-08-07
Wild3: Add cycle at moments logs and debugger tool

RainingChain:
image

Next PR is to adapt the generator cycle counter to have as few mismatch as possible with the in-game cycle counter.


zaksabeast (approved):

zaksabeast:
Sorry for the late review - I didn't look too in-depth, but it seems fine.

PR #421, opened 2026-04-04
Emerald: Painting Reseeding Retail Guide & Webtool

RainingChain:
## Painting Reseeding Retail Guide & Webtool
The webtool integrates multiple subcomponents into one large tool.
- Existing SeedToAdvances tool to select painting frame & advance.
- New PaintingTimers to create the first battle video.
- Existing Wild3 calibration tool to validate the hit painting seed.
- Existing battle video tool to update the battle video.

The tool also supports only creates a battle video without painting.

### Testing (hit target painting frame)
- Target of the RNG manipulation: 404440000 advances
- Select `0A19`
- Golbat, lvl 40, female, inner focus, careful, 114/71/65/56/79/87
- Click Update Calibration of either result.
image
image
image
image

### Testing (missed target painting frame)
- `I already know my painting frame and advance` = true
- Frame before painting = 2586
- Advances after painting = 10000
- Golbat, lvl 40, female, inner focus, careful, 114/71/65/56/79/87
- Click Update Calibration of either result.
image
image


src/rngToolsUi/gen3/wild/wild3CalibTarget.tsx
zaksabeast: Making `targetMethod` nullable on line 81 makes this cast unsafe, and the cast is unnecessary if it's not nullable.

src/rngToolsUi/gen3/battleVideo/battleVideo.tsx
```diff
@@ -95,49 +98,67 @@ const Validator = z.object({
 
 type FormState = z.infer<typeof Validator>;
 
-const initialValues: FormState = {
-  targetAdvance: 50_000,
-  isUpdatingExisting: false,
-  console: "GBA",
+const createInitialValues = (fixedData: Props["fixedData"]): FormState => ({
```
zaksabeast: We don't have to change before merging, but it might make sense to name the fixed data type:
```tsx
type FixedData = {
  targetAdvance: number;
  isUpdatingExisting: boolean;
  existingBattleVideoAdv: number;
  isAfterPainting: boolean;
  consoleType?: Gen3Console;
};

type Props = {
  fixedData: FixedData;
};
```

Then we could use `FixedData` instead of `Props["fixedData"]`.

src/rngToolsUi/gen3/wild/wild3CalibTarget.tsx
```diff
@@ -78,7 +78,7 @@ const Validator = z.object({
   targetFrameBeforePainting: z.number().min(1).max(0xffff),
   usingBattleVideoWithoutPainting: z.boolean(), // with painting, battle video is always used.
   existingBattleVideoAdv: z.number().min(1).max(0xffffffff),
-  targetMethod: z.enum(supportedGen3Methods),
+  targetMethod: z.enum(supportedGen3Methods).nullable(),
```
zaksabeast: Why is this nullable?

It looks like `createTargetSetup` in `paintingReseeding.tsx` uses `null` for `targetMethod`, but it's not clear why that's happening either.

zaksabeast (commented):
One question, but otherwise looks good

src/rngToolsUi/gen3/wild/wild3CalibTarget.tsx
RainingChain: @zaksabeast When calibrating painting, the players don't have a targetMethod. But it's probably simpler to use a dummy targetMethod, and use `isPaintingSeedConfirmed` to determine if painting is being calibrated.

PR #423, opened 2026-04-05
Add FormikEmeraldFrameBeforePaintingInput

RainingChain:


src/components/numberInput.tsx
```diff
@@ -148,3 +151,86 @@ export const FormikNumberInput = <FormState extends GenericForm>({
...
+  ];
+
+  return (
+    <Flex vertical gap={4}>
+      <RadioGroup
+        value={numType}
+        options={options}
+        onChange={(val) => {
+          setNumType(val.target.value as "hex" | "decimal");
+        }}
```
zaksabeast: Telling `RadioGroup` the correct type allows us to not cast:
```suggestion
      <RadioGroup<NumType>
        value={numType}
        options={options}
        onChange={(val) => {
          setNumType(val.target.value);
        }}
```

src/components/numberInput.tsx
```diff
@@ -148,3 +151,86 @@ export const FormikNumberInput = <FormState extends GenericForm>({
...
+        optionType="button"
+      />
+      {match(numType)
```
zaksabeast: It looks like the number input is duplicated twice. We should be able to do this:
```tsx
      <NumberInput
        status={status}
        {...props}
        numType={numType}
        name={name}
        onBlur={onBlur}
        onChange={onChange}
        value={value}
        errorMessage={errorMessage === null ? undefined : error}
      />
      {value != null &&
        match(numType)
          .with("hex", () => `Decimal: ${formatLargeInteger(value)}`)
          .with("decimal", () => `Hex: ${formatHex(value, byteCount)}`)
          .exhaustive()}
```

src/components/numberInput.tsx
```diff
@@ -148,3 +151,86 @@ export const FormikNumberInput = <FormState extends GenericForm>({
...
+                status={status}
+                {...props}
+                numType="hex"
+                name={name}
+                onBlur={onBlur}
+                onChange={onChange}
+                value={value}
+                errorMessage={errorMessage === null ? undefined : error}
+              />
+              {value != null && `Decimal: ${formatLargeInteger(value)}`}
```
zaksabeast: What's the benefit of showing `Decimal` for hex values and `Hex` for decimal values?

At a glance, it seems like if a user is typing hex they don't care about decimal, and vice versa.

src/components/numberInput.tsx
```diff
@@ -148,3 +151,86 @@ export const FormikNumberInput = <FormState extends GenericForm>({
...
+  };
+
+  const [numType, setNumType] = React.useState<NumType>(initialNumType);
+  const options = [
+    { label: "Decimal", value: "decimal" } as const,
+    { label: "Hex", value: "hex" } as const,
+  ];
+
+  return (
+    <Flex vertical gap={4}>
```
zaksabeast: Is there a reason the radio button needs to be above the input instead of next to it?

It seems like it can fit on mobile and desktop:
| Mobile | Desktop |
| - | - |
| Image | Image |

src/components/numberInput.tsx
```diff
@@ -148,3 +151,86 @@ export const FormikNumberInput = <FormState extends GenericForm>({
...
+
+type FormikNumberDecimalHexInputProps<FormState extends GenericForm> = Omit<
+  FormikNumberInputProps<FormState>,
+  "numType"
+> & {
+  initialNumType: NumType;
+  byteCount: number;
+};
+
+export const FormikNumberDecimalHexInput = <FormState extends GenericForm>({
```
zaksabeast: How come this should be a single input instead of two form fields? It feels like this could be a radio button separate from the number input.

zaksabeast (changes requested):
I'm a little confused by the use case – it feels like users should always have one type of input for a field. Why would a user have either a decimal or hex for the same input (frame before painting, in this case)?

src/components/numberInput.tsx
RainingChain: The point of that component is to avoid having to create 2 fields and a radio button for each webtool.

PR #425, opened 2026-04-11
Wild3: Improve the Wild Method Likelihood Calculator output

RainingChain:
image

How it looks without the 750px max-width:
image


src/rngToolsUi/gen3/wild/wild3CycleAtMoments.tsx
```diff
@@ -23,152 +56,496 @@ const uiCycleAtMomentColumns: ResultColumn<UiResultCycleAtMoment>[] = [
...
+      cycleAtMomentsFromTool,
+      leadCycleSpeed,
+      advanceAtSweetScent,
+    ).then((uiResult) => {
+      setUiResult(uiResult);
     });
-  }, [cycleAtMoments, compareCycleAtMomentsStr]);
+  }, [
+    compareCycleAtMomentsStr,
+    JSON.stringify(cycleAtMomentsFromTool),
```
RainingChain: It seems there's a bug somewhere because if I just put `cycleAtMomentsFromTool`, it auto-refresh infinitely.
The issue might be on the parent, but I don't really know how to investigate the issue.

zaksabeast: There are two problems:

This `useEffect` calls `setUiResult`, which triggers a re-render. Since `cycleAtMomentsFromTool` is always a new array with a new reference every time the component renders, the `useEffect` will always trigger. `cycleAtMomentsFromTool` having a stable reference solves the problem.

The second problem is mutating the prop (`cycleAtMomentsFromTool = [ /* something /* ]`), which causes funky behavior in React and also hides these types of problems.

For example, if we stop mutating the prop:
```tsx
export const Wild3CycleAtMoments = ({
  cycleAtMomentsFromTool: _cycleAtMomentsFromTool,
  leadCycleSpeed,
  advanceAtSweetScent,
}: Props) => {
  const cycleAtMomentsFromTool = [
    { moment: "SweetScentWildEncounter", cycle: 0 },
    ..._cycleAtMomentsFromTool,
  ];
```
Eslint tells us there's a problem:
Image

Removing the mutation and moving this inside `useEffect` solves the infinite rendering issue:
```tsx
React.useEffect(() => {
  const cyclesWithSweetScent: CycleAtMoment[] = [
    { moment: "SweetScentWildEncounter", cycle: 0 },
    ...cycleAtMomentsFromTool,
  ];
  createUiResult(
    compareCycleAtMomentsStr,
    cyclesWithSweetScent,
    leadCycleSpeed,
    advanceAtSweetScent,
  ).then((uiResult) => {
    setUiResult(uiResult);
  });
}, [
  compareCycleAtMomentsStr,
  cycleAtMomentsFromTool,
  leadCycleSpeed,
  setUiResult,
  advanceAtSweetScent,
]);
```

---

I've been meaning to add the [`no-param-reassign` lint rule](https://eslint.org/docs/latest/rules/no-param-reassign) to catch these types of problems. There are ~10 places variables are reassigned that could become or hide bugs in the future.

RainingChain: Ah, that makes sense. I totally forgot about `cycleAtMomentsFromTool = [ .. `

PR #426, opened 2026-04-25
Emerald Wild Guide & Webtool

RainingChain:
Emerald Wild Step-by-step Webtool. This combines multiple components that were added in previous PR into a single unified page.

Steps:
1. Find Best Setup
2. Create Battle Video (with or without Painting Reseeding)
3. Catch and Calibrate.

Steps can be skipped if the player already knows the info (ex: obtained from other tools)


src/utils/formatLargeInteger.ts
```diff
@@ -1,3 +1,7 @@
 export const formatLargeInteger = (number: number) => {
   return number.toFixed(0).replace(/\B(?=(\d{3})+(?!\d))/g, ",");
 };
+
+export const formatLargeIntegerWithSign = (number: number) => {
+  return (number >= 0 ? "+" : "") + formatLargeInteger(number);
```
zaksabeast, line 6: Should this handle `-` as well?

src/rngToolsUi/gen3/wild/wild3CalibCaughtMon.component.tsx
```diff
@@ -0,0 +1,407 @@
...
+    });
+    const sortedLvls = Array.from(lvls).sort((lvl1, lvl2) => lvl1 - lvl2);
+
+    Promise.all([
+      getStatRange({
+        species: selectedSpecies,
+        levelRange: [selectedLvl, selectedLvl],
+        nature: selectedNature,
+      }),
+    ]).then(([minMaxStats]) => {
```
zaksabeast: Any reason to use `Promise.all([getStatRange()]).then` instead of `getStatRange().then`?

src/components/abilityFilter.tsx
```diff
@@ -51,6 +51,10 @@ const getAbilityFilterOptions = async (
       ];
     }
     return [
+      {
+        label: "Any",
+        value: null,
+      },
```
zaksabeast: Seems like this should use `permitAny`

zaksabeast: There's also an `any` object a few lines up we can reuse.

RainingChain: @zaksabeast I'll split this PR in 2.

src/utils/formatLargeInteger.ts
RainingChain: `-` is automatically added when `number` is negative.

zaksabeast: Whoops, you're right

zaksabeast (approved):

PR #431, opened 2026-04-28
Auto-update FormikRadio value if it's the current value is invalid

RainingChain:
At the same time, I remove FormikGenderFilter which becomes unnecessary.


src/components/radio.tsx
```diff
@@ -14,10 +14,11 @@ import { Path, Paths } from "~/types";
 import { PrimitiveAtom, useAtom } from "jotai";
 import { Translation } from "~/translations";
 import { useActiveRouteTranslations } from "~/hooks/useActiveRoute";
+import React from "react";
+import isEqual from "lodash-es/isEqual";
 
 type RadioOptions<OptionValues extends string | number> =
-  | OptionValues[]
-  | CheckboxOptionType<OptionValues>[];
+  CheckboxOptionType<OptionValues>[];
```
RainingChain, line 21: I think only supporting CheckboxOptionType is simpler. There was only 1 use case of OptionValues[], which is easy to adapt (wrap with `toOptions`)

PR #437, opened 2026-05-02
Wild3: Improve labels, hide unsupported options (Rock Smash and Safari),

RainingChain:
New PID cycle speed input: 
image

New TooltipWithIcon:
image

Split calibration into 3 fields:
image


src/components/resultTable.tsx
```diff
@@ -91,6 +94,19 @@ export const ResultTable = <Record extends tst.O.Object>(
...
+    title: React.ReactNode,
+    tooltip?: React.ReactNode,
+  ) => {
+    if (tooltip == null) {
+      return title;
+    }
+    return (
+      <Tooltip title={tooltip}>
+        {title} <QuestionCircleOutlined />
+      </Tooltip>
```
zaksabeast, line 107: I won't hold the PR up over this, but is there a reason we shouldn't use `TooltipWithIcon` here?

PR #441, opened 2026-05-05
Wild3: Enable Rock Smash feature

RainingChain:
I messed up a bit the order of the last few PRs. This one should fix things up.


zaksabeast:
Oh whoops, I should have called this out. I noticed there were still a few rock smash todos, but figured there was a reason for it. My mistake!

PR #443, opened 2026-05-15
Wild3: Add support for all pokeblocks + display how to create them

RainingChain:
image


src/components/wild3Pokeblock.tsx
```diff
@@ -68,6 +71,37 @@ export const FormikWild3Pokeblock = <FormState extends GenericForm>({
...
+    const withWho = match(npcs)
+      .with("Npc0", () => "solo")
+      .with("Npc1", () => "with 1 NPC")
+      .with("Npc2", () => "with 2 NPCs")
+      .with("Npc3", () => "with 3 NPCs")
+      .with("BlendMaster", () => "with Blend Master")
+      .exhaustive();
+    return `Created by playing Berry Blender ${withWho}, and blending a ${berryNames[player_berry_idx]} berry.`;
+  }
+  if ("Grey" in info) {
```
zaksabeast: I wonder if this check can be more type safe.

src/components/wild3Pokeblock.tsx
```diff
@@ -68,6 +71,37 @@ export const FormikWild3Pokeblock = <FormState extends GenericForm>({
...
+    return "1/10 chance to be created by playing Berry Blender with another player, when both players provide the same berry.";
+  }
+  if ("Multiplayer" in info) {
+    const { berries } = info["Multiplayer"];
+    const berryToBlendNames = berries
+      .filter((berry) => berry < berryNames.length)
+      .map((berry) => berryNames[berry]);
+
+    return `Created by playing Berry Blender with ${berryToBlendNames.length} other players, and blending the berries: ${berryToBlendNames.join(", ")}.`;
```
zaksabeast: Would `berryToBlendNames.length` give total players? If so, do we need `${berryToBlendNames.length - 1} other players`?

Or even `Math.max(berryToBlendNames.length - 1, 0)`?

src/components/wild3Pokeblock.tsx
```diff
@@ -68,6 +71,37 @@ export const FormikWild3Pokeblock = <FormState extends GenericForm>({
   );
 };
 
+const getPokeblockCreationInfoText = (pokeblock: Pokeblock) => {
+  const key = pokeblock.join("");
+  const info = pokeblockCreationInfos.get(key);
+  if (info == null) {
+    return "";
+  }
+  if ("Npc" in info) {
```
zaksabeast: Not a requirement, but it's worth considering a stricter check here.

`"key" in obj` doesn't guarantee `key` can ever be on `obj`.

If `info` changes shape in the future, this won't be a type error, which will make it easier to introduce bugs.

Another option is to have a utility that reshapes the data with `match`:
```ts
const exists = P.not(P.nullish)
const formattedInfo = match(info)
  .with({ Npc: exists}, ({ Npc }) => ({ type: "Npc" as const, ...Npc }))
  .with({ Grey: exists}, ({ Grey }) => ({ type: "Grey" as const, ...Grey }))
  .with({ Multiplayer: exists}, ({ Multiplayer }) => ({ type: "Multiplayer" as const, ...Multiplayer }))
  .exhaustive();
```

`match` will give a type error if you try to match on a type that doesn't exist, so you can't try to access fields that aren't possible.

After that:
```ts
if (formattedInfo.type === 'Npc') {
  // logic
}
```

zaksabeast (commented):
Does `wild3MethodDistribution` need to use `canUsePokeblock && values.safariPokeblock !== null` to avoid using stale data?

src/components/wild3Pokeblock.tsx
RainingChain: Yeah, I forgot a -1 to take into consideration the "other" players.

RainingChain:
> Does `wild3MethodDistribution` need to use `canUsePokeblock && values.safariPokeblock !== null` to avoid using stale data?

In generator.rs, I do check for map_data.is_safari, but I don't do the full check based on action. It would be safer to add the canUsePokeblock check.

RainingChain:
Pushed the changes.

PR #444, opened 2026-05-16
Update to tsgo for x7 faster linting

RainingChain:


src/sw.ts
```diff
@@ -1,12 +1,19 @@
-self.addEventListener("fetch", (event: FetchEvent) => {
-  event.respondWith(
+type ServiceWorkerFetchEvent = Event & {
+  request: Request;
+  respondWith: (response: Promise<Response>) => void;
+};
+
+self.addEventListener("fetch", (event) => {
```
RainingChain, line 6: Proper fix would require messing around with `compilerOptions.lib`. Didn't manage to make it work...

zaksabeast: This works well enough. I've been meaning to redo this worker anyways.

PR #445, opened 2026-05-17
Swarm

fiaskotto:
Add `swarm_filter` for Gen 4 swarm RNG

Adds a new function that, given a seed and an advance range, returns all advances where a specific swarm route appears.

The logic mirrors `generate_swarm` in `swarm.rs` but filters by a wanted route instead of returning everything. Uses the existing `MT`, `SwarmRoute`, `DPPT_ROUTES` and `HGSS_ROUTES` from the same module, no new dependencies. (`DPPT_ROUTES` and `HGSS_ROUTES` became `pub` because they are necessary for `swarm_filter`)

The user inputs a seed (found via the existing seed time tools), selects a target route and an advance range, and gets back the list of frames where that route is in swarm.


rng_tools/src/generators/gen4/swarm_filter.rs
zaksabeast, line 1: Let's either replace the contents of `swarm.rs` with the contents of this file, or remove `swarm.rs`. There's no need to keep both.

Either way, we'll need to update `rng_tools/src/generators/gen4/mod.rs` to either add this module or remove `swarm.rs`.

rng_tools/src/generators/gen4/swarm_filter.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+//Input
+#[derive(Debug, Clone, Copy, PartialEq, Tsify, Serialize, Deserialize)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub struct SwarmFinderOpts {
+    pub seed: u32,
+    pub min_advances: usize,
+    pub max_advances: usize,
+    pub wanted_route: SwarmRoute,
```
zaksabeast, line 21: Let's make `wanted_route` optional, in case a user wants to see all routes for some reason:
```rs
pub wanted_route: Option<SwarmRoute>,
```

Then below we can do this:
```rs
let res = SwarmFinderResult {
    advance: opts.min_advances + i,
    route,
};
match opts.wanted_route {
    // If we don't have a filter, return the current route
    None => Some(res),
    // If we have a filter that matches the current route, return it
    Some(wanted) if wanted == route => Some(res),
    // If we have a filter that doesn't match the current route, skip it
    _ => None,
}
```

rng_tools/src/generators/gen4/swarm_filter.rs
```diff
@@ -0,0 +1,60 @@
...
+                Some(SwarmFinderResult {
+                    advance: opts.min_advances + i,
+                    route,
+                })
+            } else {
+                None
+            }
+        })
+        .collect()
+}
```
zaksabeast, line 60: Let's add tests to the bottom of this file the same way `swarm.rs` does.

We should have three tests:
1. Unfiltered dppt routes match expected values
2. Unfiltered hgss routes match expected values
3. Filter routes for one or both games match expected values

That way if we make changes in the future and `cargo test` passes, we know everything works as expected.

zaksabeast (commented):
Nice! Just a few tiny tweaks and we should be good to go.

rng_tools/src/generators/gen4/swarm.rs
```diff
@@ -139,72 +225,226 @@ fn generate_swarm(opts: &SwarmOpts, routes: &[SwarmRoute]) -> Vec<SwarmEncounter
...
-pub fn generate_dppt_swarm(opts: &SwarmOpts) -> Vec<SwarmEncounter> {
-    generate_swarm(opts, &DPPT_ROUTES)
+pub fn generate_dp_swarm(opts: &SwarmOpts) -> Vec<SwarmEncounter> {
+    generate_swarm(opts, &DP_ENCOUNTERS)
+}
+
+#[wasm_bindgen]
+pub fn generate_pt_swarm(opts: &SwarmOpts) -> Vec<SwarmEncounter> {
+    generate_swarm(opts, &PT_ENCOUNTERS)
+}
```
zaksabeast: Let's remove the `generate_<game>_swarm` and `generate_swarm` since `find_swarm_advances` can generate Pokemon without a filter.

zaksabeast (changes requested):
Two more small things!

Aside from the other comment, let's remove the `rng_tools/src/generators/gen4/swarm_filter.rs` file since it isn't being used anymore.

Once those are done, this is good to merge.

zaksabeast:
Whoops, one final thing: can you run `cargo fmt` inside the `rng_tools` directory? That will format the code to be consistent with the rest of the code.

You can also set up VSCode to automatically format code!

PR #447, opened 2026-05-21
Wild3: Improve performance

RainingChain:
- Removed most dynamic memory allocation by replacing Vec by ArrayVec.
- Removed most intermediate vectors
- Added ReversePidShiny searching strategy
- Replaced sort().take() by k_smallest
- Considered methods are now const (generic argument for zero-overhead).


rng_tools/src/generators/gen3/wild/searcher/tests/searcher_reverse_perf_tests.rs
```diff
@@ -123,3 +123,139 @@ fn test_search_reverse_perf_vanilla() {
...
+    let opts = FindPidPathsOptions {
+        filter: PkmFilter {
+            shiny: true,
+            min_ivs: Ivs::new(0, 0, 0, 0, 0, 0),
+            max_ivs: Ivs::new(20, 20, 20, 20, 20, 20), // ~85M
+            ..Default::default()
+        },
+        methods: vec![Gen3Method::Wild1],
+        tsv: 1234,
+        forced_search_strategy: Some(PidPathStrategy::ReverseIv),
```
zaksabeast: It doesn't look like `find_pid_paths_reverse_iv` ever uses `forced_search_strategy`. Does that not matter for this test?

`forced_search_strategy` is used by `determine_best_pid_path_strategy`, and the only call path seems to be `search_wild3_reverse` -> `search_wild3_reverse_with_methods` -> `determine_best_pid_path_strategy`.

rng_tools/src/generators/gen3/wild/searcher/pid_low_path.rs
```diff
@@ -61,50 +58,69 @@ pub enum PidLowToIvArc {
...
+    while i < methods_to_check.len() {
+        let method = methods_to_check[i];
+        if method >= 1 && method <= 4 && (opts_methods & (1 << (method - 1))) != 0 {
+            return true;
+        }
+        i += 1;
+    }
+
+    false
+}
```
zaksabeast: Is there a reason to use a slice instead of a bitmask?

```rs
const METHOD_1: u8 = 0b0001;
const METHOD_2: u8 = 0b0010;
const METHOD_3: u8 = 0b0100;
const METHOD_4: u8 = 0b1000;

pub const fn is_considered_method(opts_methods: u8, methods_to_check: u8) -> bool {
    opts_methods & methods_to_check != 0
}
```

That would also make it a little easier to read the generics:
```rs
is_considered_method(METHODS, METHOD_1 | METHOD_2 | METHOD_4)

find_pid_paths_by_step_pid::<{ METHOD_1 | METHOD_2 }>(&opts)
```

rng_tools/src/generators/gen3/wild/searcher/searcher_reverse.rs
```diff
@@ -194,57 +187,79 @@ fn new_find_pid_paths_options(opts: &Wild3SearcherOptions) -> FindPidPathsOption
...
+    }
+    if opts.methods.contains(&Gen3Method::Wild2) {
+        methods_bits += 1 << 1;
+    }
+    if opts.methods.contains(&Gen3Method::Wild3) {
+        methods_bits += 1 << 2;
+    }
+    if opts.methods.contains(&Gen3Method::Wild4) {
+        methods_bits += 1 << 3;
+    }
```
zaksabeast, line 222: It looks like this would give incorrect results if we accidentally provided duplicate methods. Providing `Gen3Method::Wild1` twice would have `method_bits` at 2, not 1.

If we used the method consts from the previous comment, we could prevent duplicate issues with this:
```rs
if opts.methods.contains(&Gen3Method::Wild1) {
    methods_bits |= METHOD_1;
}
```

rng_tools/src/generators/gen3/wild/searcher/searcher_reverse.rs
```diff
@@ -194,57 +187,79 @@ fn new_find_pid_paths_options(opts: &Wild3SearcherOptions) -> FindPidPathsOption
...
+
+    match methods_bits {
+        0 => search_wild3_reverse_with_methods::<0>(opts),
+        1 => search_wild3_reverse_with_methods::<1>(opts),
+        2 => search_wild3_reverse_with_methods::<2>(opts),
```
zaksabeast, line 227: If we use method consts, this can remain the same. This is cleaner than mapping `search_wild3_reverse_with_methods` to all method consts and the intent is already clear.

zaksabeast (changes requested):
Looks like there's one tiny bug, and I had a few questions.

rng_tools/src/generators/gen3/wild/searcher/searcher_reverse.rs
RainingChain: The original code doesn't have the problem you mention, because even if Wild1 is included multiple times, `if opts.methods.contains(&Gen3Method::Wild1) {` is only called once.
But anyway, I'll do your proposed change.

zaksabeast: You're right, in my haste I misread this as part of a loop

PR #449, opened 2026-05-23
Wild3: Improve performance of searching for specific PID cycle speed

RainingChain:


rng_tools/src/generators/gen3/wild/most_dense_pid_chunks_by_spd.rs
zaksabeast, line 1: @RainingChain I'm going to merge this PR so the functionality goes live, but should this file go in a `tests` directory?

RainingChain: @zaksabeast It's not really a test. It's the code for calculating/updating MOST_DENSE_PID_CHUNKS_BY_SPD.
In an ideal world, MOST_DENSE_PID_CHUNKS_BY_SPD would be calculated on the fly, but it would take like 2 minutes to do so. I suggest leaving that code there.

PR #450, opened 2026-05-24
Improve PkmFilter to support multiple natures + improve labels

RainingChain:
image

image


rng_tools/src/generators/gen3/wild/searcher/searcher_naive.rs
```diff
@@ -38,14 +38,9 @@ fn search_wild3_naive_at_given_advance(
...
 
-                if let Some(nature) = opts.filter.nature {
-                    vec![None, to_pokeblock(nature)]
-                } else {
-                    (0..NATURE_COUNT)
-                        .map(|i| to_pokeblock((i as u8).into()))
-                        .chain(std::iter::once(None))
-                        .collect_vec()
-                }
+                std::iter::once(None)
```
zaksabeast: Why is this needed instead of using `permitted_natures_iter` directly?

rng_tools/src/pkm/mod.rs
```diff
@@ -36,7 +36,7 @@ use tsify::Tsify;
 #[tsify(into_wasm_abi, from_wasm_abi)]
 pub struct PkmFilter {
     pub shiny: bool,
-    pub nature: Option<Nature>,
+    pub nature: Option<[bool; NATURE_COUNT]>,
```
zaksabeast, line 39: Is a list of bools used to prevent duplicates, or is it for performance?

A list of bools is effectively a bitmask. Is there a reason to not use a bitmask? (e.g. set with `nature_filter |= (1 << nature)` and get with `nature_filter & (1 << nature)`)

src/components/pkmFilter.tsx
```diff
@@ -21,45 +27,176 @@ import {
...
+      labelWhenSelected: t("Sassy"),
+      label: toNatureLabel(t("Sassy"), t("SpD"), t("Spe")),
+    },
+    { value: "Serious", labelWhenSelected: t("Serious"), label: t("Serious") },
+    {
+      value: "Timid",
+      labelWhenSelected: t("Timid"),
+      label: toNatureLabel(t("Timid"), t("Spe"), t("Atk")),
+    },
+  ] as { value: Nature; labelWhenSelected: string; label: React.ReactNode }[];
```
zaksabeast: Let's use `satisfies` instead of `as`:
```ts
satisfies {
    value: Nature;
    labelWhenSelected: string;
    label: React.ReactNode;
  }[]
```

src/components/pkmFilter.tsx
```diff
@@ -21,45 +27,176 @@ import {
...
-  [Key in keyof PkmFilter as `filter_${Key}`]: undefined extends PkmFilter[Key]
-    ? tst.U.Exclude<PkmFilter[Key], undefined> | null
-    : PkmFilter[Key];
+  filter_shiny: boolean;
+  filter_nature: Nature[];
+  filter_gender: Gender | null;
+  filter_min_ivs: Ivs;
+  filter_max_ivs: Ivs;
+  filter_ability: AbilityType | null;
+  filter_hidden_power: HiddenPowerFilter;
```
zaksabeast, line 179: Let's ground in some type of truth from rust to make sure we don't forget to edit this if needed.

We can create a file at `src/components/__tests__/pkmFilter.type-test.ts` with these contents:
```ts
import { check, Pass } from "~/typeTest";
import { type PkmFilterFields } from "../pkmFilter";
import { type PkmFilter } from "~/rngTools";

export const PkmFilterMatchesRust = () => {
  type Result = keyof PkmFilterFields;
  type Expected = `filter_${keyof PkmFilter}`;
  check<Result, Expected>(Pass);
};
```
and it'll give a type error if field names don't match.

src/components/natureFilterInput.tsx
```diff
@@ -0,0 +1,41 @@
...
+    } else {
+      lastSpecificNatures.current = selectedNatures;
+      setNature([]);
+    }
+  };
+
+  return (
+    <Flex gap={20} vertical>
+      <Flex vertical align="start">
+        <Switch value={active} onChange={updateIsSpecific} />
```
zaksabeast: Is there a reason to have a switch? It seems like the user could leave the field empty if they don't want to filter natures.

We could have a `placeholder` to help show empty means all natures:
Image

src/components/pkmFilter.tsx
```diff
@@ -21,45 +27,176 @@ import {
...
 import { FormikAbilityFilter } from "./abilityFilter";
 import { FormikRadio } from "./radio";
+import { NatureFilterInput } from "./natureFilterInput";
+import { Flex } from "./flex";
+import { Icon } from "./icons";
 
-const sortedNatures = nature.toSorted();
+const toNatureLabel = (nat: string, up: string, down: string) => {
+  return (
+    <Flex>
```
zaksabeast: It's a little cramped on mobile after selecting a nature.

Maybe we add `mr={8}` here?

| No margin | With mr={8} |
| - | - |
| Image | Image |

src/components/natureFilterInput.tsx
```diff
@@ -0,0 +1,41 @@
...
+      </Flex>
+      {active && (
+        <FormikSelect<PkmFilterFields, "filter_nature">
+          name="filter_nature"
+          mode="multiple"
+          options={getTranslatedNatureOptions()}
```
zaksabeast: Seems like we should add this to the top of the NatureFilterInput:
```tsx
const t = useActiveRouteTranslations();
```
then pass it to the options:
```tsx
options={getTranslatedNatureOptions(t)}
```

zaksabeast (changes requested):
This is really cool! Very helpful update, thank you!

---

It looks like typing to filter selects seems to not be working. Can you reproduce this, or is it just me?

| Main | This PR |
| - | - |
| Image | Image |

rng_tools/src/pkm/mod.rs
RainingChain: @zaksabeast 
It's for performance.
`[bool;25]` is faster than a bitmask, because it doesn't need to do bit shifting.
I made a quick benchmark for that
[bool;25] indexing: 43.5274ms
u32 bitmask indexing: 46.0887ms

Using `[bool;25]` is also easier to understand.

```
// cargo test --release compare_bool_array_and_u32_bitmask_indexing --lib -- --ignored --nocapture
#[cfg(test)]
mod benchmarks {
    use std::{
        hint::black_box,
        time::{Duration, Instant},
    };

    const VALUE_COUNT: usize = 25;
    const ITERATIONS: usize = 50_000_000;
    const MASK: u32 = 0x0155_5555;

    #[inline(always)]
    fn array_value(values: &[bool; VALUE_COUNT], index: usize) -> bool {
        values[index]
    }

    #[inline(always)]
    fn bitmask_value(mask: u32, index: usize) -> bool {
        ((mask >> index) & 1) != 0
    }

    fn benchmark<F>(mut get_value: F) -> Duration
    where
        F: FnMut(usize) -> bool,
    {
        let start = Instant::now();
        let mut matches = 0_usize;

        for iteration in 0..ITERATIONS {
            let index = black_box(iteration % VALUE_COUNT);
            matches += get_value(index) as usize;
        }

        black_box(matches);
        start.elapsed()
    }

    #[test]
    #[ignore = "benchmark"]
    fn compare_bool_array_and_u32_bitmask_indexing() {
        let values = std::array::from_fn(|index| bitmask_value(MASK, index));

        let array_duration = benchmark(|index| array_value(black_box(&values), index));
        let bitmask_duration = benchmark(|index| bitmask_value(black_box(MASK), index));

        eprintln!("[bool;25] indexing: {array_duration:?}");
        eprintln!("u32 bitmask indexing: {bitmask_duration:?}");
    }
}
```

src/components/pkmFilter.tsx
RainingChain: The actual point of that change is to separate typing from rust and typing from TS. Rust typing should be optimized for speed, while TS typing should be optimized for UI simplificity.

For nature, rust typing is [bool;25] while TS typing is Nature[]. I also plan to do the same with hidden power typing in another PR.

There is no need for an explicit test. `bun run check` already detects issue if `pkmFilterFieldsToRustInput` is incomplete.

zaksabeast: The type test only makes sure the keys match so we're not missing any new keys or renames. Individual types for each field aren't carried over, so rust and TS can have separate implementation.

But you're right that `pkmFilterFieldsToRustInput` should catch that for us. Before our input fields had type safety, validation, and common utils, we had a few bugs with TS inputs not being formatted correctly for rust. Derived types helped prevent problems while the foundation for the web tools were being built. But at this point, it's not needed anymore.

In fact, we should probably do this:
```ts
export type PkmFilterFields = z.infer<typeof pkmFilterSchema>;
```

rng_tools/src/pkm/mod.rs
zaksabeast: Here are my averages after running that test a few hundred times:
- [bool;25] avg: 39.5747ms
- bitmask avg: 38.8559ms

I'd expect a memory read to take more time than common math operations, and it seems like bitmask is slightly faster.

If we try running this test in wasm over thousands of iterations, I'd expect the bitmask to be even faster.

I think both are easy to understand, so I'm good either way.

zaksabeast:
Showing nature stat effects is such a nice improvement.

PR #454, opened 2026-05-27
Permit unselecting row in ResultTable

RainingChain:
At most 1 row can be selected.

image

### DIFF
--- a/src/components/resultTable.tsx
+++ b/src/components/resultTable.tsx
@@ -86,12 +86,15 @@ const TABLE_SCROLL = { x: true } as const;
 type FormikResultTableProps<Record extends tst.O.Object> = tst.O.Overwrite<
   TableProps<Record>,
   { columns: ResultColumn<Record>[] }
->;
+> & {
+  onClickResultRow?: (record: Record | null) => void;
+};
 
 export const ResultTable = <Record extends tst.O.Object>(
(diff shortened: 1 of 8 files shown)


src/components/resultTable.tsx
```diff
@@ -166,12 +169,28 @@ export const ResultTable = <Record extends tst.O.Object>(
           ),
         };
 
+  const rowSelection: FormikResultTableProps<Record>["rowSelection"] =
+    props.rowSelection != null
+      ? props.rowSelection
```
zaksabeast: Currently every table shows a checkbox:
Image

We need to only use the checkbox when `onClickResultRow` exists:
```ts
const rowSelection: FormikResultTableProps<Record>["rowSelection"] =
  props.rowSelection ??
  (props.onClickResultRow != null
    ? {
        type: "checkbox",
        selectedRowKeys,
        hideSelectAll: true,
        onChange: (newSelectedRowKeys, selectedRows) => {
          const latestKey = newSelectedRowKeys.slice(-1);
          const latestRows = selectedRows.slice(-1);
          setSelectedRowKeys(latestKey);
          props.onClickResultRow?.(latestRows[0] ?? null);
        },
      }
    : undefined);
```

zaksabeast (changes requested):

PR #456, opened 2026-06-01
Swarm Essential tools finished

fiaskotto:
Added:
- A counter in DPPt that tracks how many coin flips have been made, and passes this value to the next step.
- In Step 4 for DPPt, a filter that shows the current advance in real time and how many flips are needed to reach the target.
- In Step 4 for HGSS, display how many more advances remain.

All that's missing is text to inform the user that Joey advances by 2 while forming a group of 1.


src/rngToolsUi/gen4/shared/dpptCoinFlip/seedFinder.tsx
```diff
@@ -55,16 +55,29 @@ type SelectButtonProps = {
...
 
   return (
     <Button
       trackerId="coin_flip_select_seed"
       onClick={() => {
+        console.log(
+          "Saving coin flips:",
+          state.gameState.coinFlips,
+          state.gameState.coinFlips.length,
+        );
```
zaksabeast: Whoops, looks like this was left by mistake. Let's remove this.

src/rngToolsUi/gen4/shared/hgssElmCalls/utils.ts
```diff
@@ -1,22 +1,46 @@
-import type { ElmCall } from "~/rngTools";
+import { type ElmCall } from "~/rngTools";
+import { findSubArrayIndices, type IndexRange } from "~/utils/findIndexBy";
 import { matchesSubsequence } from "~/utils/matchesSubsequence";
 
-const isElmCall = (char: string): char is ElmCall => {
-  return char === "E" || char === "K" || char === "P";
+export type SmallElmCall = "E" | "K" | "P";
```
zaksabeast: It looks like `ElmCall` and `SmallElmCall` are duplicate types.

Coin Flip goes from Heads and Tails to H and T, so those are shrunk, but elm calls are fine by themselves.

src/rngToolsUi/gen4/swarm/hgssSwarmAdvanceTracker.tsx
```diff
@@ -0,0 +1,106 @@
+import React from "react";
+import { useAtom } from "jotai";
+import { gen4StateAtom } from "../shared/state";
+import { Alert, Button, Flex, Typography } from "~/components";
+
+export const HgssSwarmAdvanceTracker = () => {
+  const [state] = useAtom(gen4StateAtom);
+
+  const [currentAdvance, setCurrentAdvance] = React.useState(0);
```
zaksabeast, line 9: At this point users will have used elm calls to figure out their seed, so we'll probably want to take the elm calls into account as well.

If we rename `currentAdvance, setCurrentAdvance` to `manualAdvances, setManualAdvances`, then we can calculate the `currentAdvance` with elm calls:
```ts
  const [manualAdvances, setManualAdvances] = React.useState(0);
  const elmCallCount = state.gameState.elmCalls.length;
  const currentAdvance = manualAdvances + elmCallCount;
```

src/rngToolsUi/gen4/swarm/hgssSwarmAdvanceTracker.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+        <Button
+          trackerId="hgss_swarm_advance_group"
+          disabled={targetAdvance == null || reached}
+          onClick={() => addAdvance(1)}
+        >
+          Formed Group
+        </Button>
+
+        <Button
```
zaksabeast, line 97: Right now `bun run lint:ts:types` fails due to a type error here. Let's add a `trackerId` to resolve the type error.

src/rngToolsUi/gen4/swarm/showIf.tsx
```diff
@@ -0,0 +1,24 @@
...
+  game === "Diamond" || game === "Pearl" || game === "Platinum";
+
+export const SwarmShowIf = ({ game, children }: SwarmShowIfProps) => {
+  const [state] = useSwarmState();
+
+  if (game === "dppt" && !isDppt(state.game)) {
+    return null;
+  }
+
+  if (game === "hgss" && isDppt(state.game)) {
```
zaksabeast: Looks like we have two type errors here since `game` doesn't exist on the swarm state anymore.

One option we can do is swap `useSwarmState` to:
```ts
const [state] = useAtom(gen4StateAtom);
```
then get the game with `state.config.game`.

src/rngToolsUi/gen4/shared/hgssElmCalls/elmCallButtons.tsx
zaksabeast, line 1: It looks like this file is a duplicate of `elmCallFilterButtons`. Let's remove either this file or `elmCallFilterButtons`.

`elmCallFilterButtons` is used by `seedFinder` and `elmCallButtons` is used by `advanceFilter`, so we'll need to update either `seedFinder` or `advanceFilter` as well.

src/rngToolsUi/gen4/shared/hgssElmCalls/advanceFilter.tsx
zaksabeast, line 1: It looks like we never use this file. Was this originally intended to be used, but then replaced with `HgssSwarmAdvanceTracker`?

src/rngToolsUi/gen4/swarm/findEncounter.tsx
```diff
@@ -89,6 +94,14 @@ const getColumns = (t: Translations): ResultColumn<Result>[] => [
 export const SwarmFindEncounter = () => {
   const [swarmState] = useSwarmState();
   const [state] = useAtom(gen4StateAtom);
+
+  const defaultMinAdvance =
+  state.config.game === "Diamond" ||
+  state.config.game === "Pearl" ||
+  state.config.game === "Platinum"
+    ? state.target.coinFlipCount
```
zaksabeast: Let's use `state.gameState.coinFlips.length` here and remove `coinFlipCount` from the shared state.

That way we don't have to remember to set `coinFlipCount` in the seed finder.

src/rngToolsUi/gen4/shared/hgssElmCalls/advanceFilter.tsx
```diff
@@ -0,0 +1,162 @@
...
+    title: t["Target/Seen"],
+    dataIndex: "status",
+    render: renderAdvanceFilterStatus,
+  },
+  {
+    title: t["Advance"],
+    dataIndex: "advance",
+  },
+  {
+    title: t["Elm Calls"],
```
zaksabeast: `bun run lint:ts:types` shows this isn't an available translation right now.

Let's add this to the end of `src/translations/en/index.ts`:
```
"Elm Calls": "Elm Calls",
```

guides/Gen 4/Swarm.mdx
```diff
@@ -76,25 +73,30 @@ List of calls in HeartGold and SoulSilver:
...
 </Step>
 
 <Step step={3}>
 
-## Get the Pokémon
+## Advance the RNG and Get the Pokémon
 
-1. If you are on Diamond, Pearl, or Platinum, take into account all coin flips already performed.
-2. Do whatever is necessary to reach your MT Advance Target.
-3. Once you have reached your target, talk to the group NPC and make a new group (in DPPt it is standing in Jubilife City in front of the Jubilife TV Station, in HGSS it is inside the Pokémon School of Violet City)
```
zaksabeast, line 97: Casual players won't know what a group NPC is. We might want to let users know this is the NPC they are saved in front of.

zaksabeast (changes requested):
Just a few small tweaks, but otherwise this is looking good!

src/rngToolsUi/gen4/swarm/hgssSwarmAdvanceTracker.tsx
zaksabeast: Whoops, Elm is LCRNG! Ignore this!

src/rngToolsUi/gen4/shared/hgssElmCalls/utils.ts
```diff
@@ -1,22 +1,44 @@
-import type { ElmCall } from "~/rngTools";
+import { type ElmCall } from "~/rngTools";
+import { findSubArrayIndices, type IndexRange } from "~/utils/findIndexBy";
 import { matchesSubsequence } from "~/utils/matchesSubsequence";
 
-const isElmCall = (char: string): char is ElmCall => {
-  return char === "E" || char === "K" || char === "P";
+export const shrinkElmCalls = (elmCalls: ElmCall[]): ElmCall[] => {
```
zaksabeast: Looks like we can remove this function since nothing is using it.

guides/Gen 4/Swarm.mdx
```diff
@@ -76,25 +73,30 @@ List of calls in HeartGold and SoulSilver:
...
+1. Use the tool below to keep track of your current advance.
+2. Advance until you are on the right advance. (The one with the checkmark)
+3. Once you have reached your target, talk to the group NPC and make a new group.
```
zaksabeast, line 98: Should we mention the group NPC is the NPC the user is saved in front of?

zaksabeast (changes requested):
Almost there!

A few commands need to be run:
- `build:guides`: Since a change was made to an mdx file, we need to run this to rebuild the generated guide data.
- `bun run lint:ts`: Shows a few tiny code changes that need to be made, and file formatting needs to be updated.
- `bun run format` will automatically format all files so you don't have to manually.

`CONTRIBUTING.md` has a "Automatic formatting" section if you don't want to run `bun run format` all the time!

PR #458, opened 2026-06-07
Gen 3 Static RNG (Finding Target)

RainingChain:
The first PR for Gen 3 Static RNG.
Next PR will add calibration for Emerald.

image
image


src/rngToolsUi/gen3/paintingReseeding/paintingCache.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  const advsWithoutDupe = new Uint32Array(new Set(advs));
+  const fastestAdvs = await rngTools.find_fastest_advs_considering_painting(
+    opts,
+    advsWithoutDupe,
+  );
+
+  if (fastestAdvs.length !== advsWithoutDupe.length) {
+    return fallbackFunc;
+  }
```
zaksabeast: Nothing to change, but for my own knowledge: why do we use the fallback func instead of using the de-duped advances?

src/rngToolsUi/gen3/static/static3TargetSetupSearcher.tsx
```diff
@@ -0,0 +1,123 @@
...
+      validationSchema={schema}
+      initialValues={initialValues}
+      onSubmit={onSubmit}
+      submitTrackerId="wild3_find_target"
+      rowKey="uid"
+      onClickResultRow={onClickResultRow}
+    >
+      <Static3TargetMon game={game} />
+      <br />
+      <Static3SetupFilter game={game} />
```
zaksabeast: Tiny nit: let's use a `Flex` here with a `gap`.

`br` works well for text, but any future global style changes we make for text should not affect the layout of non-text components.

src/rngToolsUi/gen3/static/static3TargetMon.tsx
```diff
@@ -0,0 +1,113 @@
...
+  FormikSwitch,
+  Flex,
+  Typography,
+  Link,
+} from "~/components";
+import { getGen3PkmFilterFields } from "~/components/gen3PkmFilter";
+import { getPkmFilterFields } from "~/components/pkmFilter";
+import { Species } from "~/rngTools";
+import { toOptions } from "~/utils/options";
+import { useWatch } from "react-hook-form";
```
zaksabeast: Let's use `useWatch` from `~/hooks/form`.

`react-hook-form`'s `useWatch` has two problems:
- It doesn't run validation before forwarding the value to us
- The types don't account for missing data (e.g. data hasn't been provided)

These previously broke type assumptions and caused a page crash.  Our version fixes those problems.

When I'm available, I can add a lint rule to catch this in the future.

src/rngToolsUi/gen3/static/static3SetupFilter.tsx
```diff
@@ -0,0 +1,85 @@
+import { gen3StaticMethods, Static3Game } from "./constants";
+import { Field, FormFieldTable, FormikSelect, Typography } from "~/components";
+import { toOptions } from "~/utils/options";
+import { FormState } from "./static3TargetSetupSearcher";
+import { useWatch } from "react-hook-form";
```
zaksabeast: Let's use `useWatch` from `~/hooks/form` here too.

package.json
```diff
@@ -18,6 +18,7 @@
     "build:guides:force": "bun run scripts/reset-guides.ts && bun run scripts/build-guides.tsx --no-detected-tags && bun run build:guides",
     "build:guides:check": "bun run scripts/build-guides.tsx --check",
     "build:rng_tools": "wasm-pack build rng_tools",
+    "build:rng_tools:dev": "wasm-pack build rng_tools --dev",
```
zaksabeast, line 21: My main concern is this will slow down testing the UI. I'm noticing slightly slower speeds when testing the app locally.

Normally we'd want to validate rust behavior with tests and UI behavior separately, so we shouldn't have to make wasm builds often.

If this is for compilation speeds, we're trading a few wasm builds for frequent UI slowdown.

Is this for debug assertions, compilation speed, or something else? Depending what we're trying to solve, maybe there's another solution?

src/rngToolsUi/gen3/static/static3TargetSetupSearcher.tsx
```diff
@@ -0,0 +1,123 @@
...
+  };
+
+  return (
+    <RngToolForm<FormState, PidPathResult>
+      columns={pidPathColumns}
+      results={pidPathResults}
+      validationSchema={schema}
+      initialValues={initialValues}
+      onSubmit={onSubmit}
+      submitTrackerId="wild3_find_target"
```
zaksabeast: Let's update this to `static3_find_target`.

src/rngToolsUi/gen3/static/searchStatic3Target.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  const searchOpts = {
+    initial_seed,
+    tid: opts.tid,
+    sid: opts.sid,
+    initial_advances,
+    max_advances,
+    max_result_count: opts.max_result_count,
+    filter: pkmFilterFieldsToRustInput(opts),
+    gen3_filter: gen3PkmFilterFieldsToRustInput(opts, opts.species),
```
zaksabeast, line 58: The wild tool overrides the shiny filter by taking `usingAceForSid` into account. Do we not want to do the same thing here at the moment?

zaksabeast (changes requested):

src/rngToolsUi/gen3/paintingReseeding/paintingCache.ts
RainingChain: Fundamentally, the issue is that if the lengths don't match, an error occured while calculating the values. I guess it doesn't really matter. I can change it.

package.json
RainingChain: On my laptop, `wasm-pack build rng_tools` takes a very long time. We could keep the default compil unchanged, and only  add`build:rng_tools:dev`. 
On my laptop, I'd use `bun run build:rng_tools:dev` and `bun run start`.
What do you think?

zaksabeast: Sounds good to me!

PR #459, opened 2026-06-10
Emerald Static Webtool + Guide

RainingChain:
Adds the battle video and calibration steps to the Emerald Static guide.
Adds basic instructions. Instructions about specific species will be in another PR.

A few minor changes to the wild guide: target rng state can now be either advances or seed. Rare candies columns now show the level.


guides/Gen 3/Emerald/Gen 3 Static.mdx
```diff
@@ -22,14 +21,90 @@
...
+
+<ShowIf slug="/frlg-static">
+  <Static3TargetSetupSearcher game="frlg" />
+</ShowIf>
+
+<ShowIf slug="/emerald-static">
+
+## General Overview
+
+Static Pokémon are starter, legendary, fossil, gift, stationary, and roamer Pokémon. Manipulating static encounters only requires 1 frame-perfect input. Calibrating is fast and reliable.
```
zaksabeast: Not a required change, just a thought

If we have to specify what "static" means, I wonder if there's a better guide name than "static".

People new to RNG won't know what "static" means, and the description is inside the guide, so you have to click into it to know what it is.

src/rngToolsUi/gen3/pokemonRng/shared.tsx
zaksabeast, line 1: I'm a little surprised by this file name. It doesn't export a component, nor does this seem like it's a main component for the RNG. This seems like it's ace state.

Should this be named something else?

src/rngToolsUi/gen3/static/static3Calib.tsx
```diff
@@ -0,0 +1,232 @@
...
+type Props = AllOrNone<{
+  targetSetup: TargetSetup;
+  battleVideoInfo: BattleVideoInfo;
+  clearAll?: () => void;
+  displayInstructions?: boolean;
+}>;
+
+export const Static3Calib = ({
+  targetSetup: targetSetupProp,
+  battleVideoInfo: battleVideoInfoProp,
```
zaksabeast, line 37: What are thoughts on prefixing these with `initial` to avoid confusion.

Since `targetSetup` and `battleVideo` can change internally, these don't represent the actual setup or video, just the initial values.

src/rngToolsUi/gen3/static/static3Calib.tsx
```diff
@@ -0,0 +1,232 @@
...
+  clearAll,
+  displayInstructions = true,
+}: Props) => {
+  const [targetSetup, setTargetSetup] = React.useState<TargetSetup | null>(
+    targetSetupProp ?? null,
+  );
+
+  React.useEffect(() => {
+    setTargetSetup(targetSetupProp ?? null);
+  }, [targetSetupProp]);
```
zaksabeast: I think we use this same pattern in a few places. Should we abstract it into a hook?

zaksabeast: One issue to be aware of: if `targetSetupProp` or `battleVideoInfoProp` has a non-stable reference, this will always run on render and cause the component to look permanently stuck on the initial values.

We might want to exit early for safety (even if we currently use stable refs):
```ts
  React.useEffect(() => {
    if (isEqual(targetSetup, targetSetupProp)) {
      return;
    }

    setTargetSetup(targetSetupProp ?? null);
  }, [targetSetupProp]);
```

`isEqual` from `lodash` handles object comparison the way we'd want for this case:
```ts
import { isEqual} from 'lodash-es'

{ a: 'test' } === { a: 'test' }
// false

isEqual({ a: 'test' }, { a: 'test' })
// true

isEqual({ a: 'test', sub: { b: 0 } }, { a: 'test', sub: { b: 0 }  })
// true
```

src/rngToolsUi/gen3/static/static3CalibCaughtMon.tsx
```diff
@@ -0,0 +1,289 @@
...
+  FormState,
+  getCommonFieldInputs,
+  initialValues,
+  updateResultsForRareCandy,
+  validator,
+} from "../pokemonRng/calibCaughtMon";
+import { getStaticEncounterLvl } from "./calculateTargetSetupResult";
+import { RngToolSubmit } from "~/components/rngToolForm";
+import { Typography } from "~/components/typography";
+import { useWatch } from "react-hook-form";
```
zaksabeast: Let's use `useWatch` from `~/hooks/form` (sorry, I haven't had the chance to add the lint rule yet)

src/rngToolsUi/gen3/static/static3CalibCaughtMon.tsx
```diff
@@ -0,0 +1,289 @@
...
+
+  const onRemove = (values: CaughtMonResult) => {
+    setResults(results.filter((res) => res !== values));
+  };
+
+  const onUpdateCalib =
+    setLatestHitAdv == null
+      ? undefined
+      : (values: CaughtMonResult) => {
+          setLatestHitAdv?.(values.advance, values.method as Gen3StaticMethod);
```
zaksabeast: Let's remove the type cast. The TS config will be updated at some point to prohibit most casting and only allow a small handful of exceptions.

One option is to make `method` generic in `CaughtMonResult`, then have aliases for `type CaughtStatic3MonResult = CaughtMonResult<Gen3StaticMethod>` and another for `CaughtMonResult<Gen3Method>`.

Another option is to wrap this with a check:
```ts
const isType = <T extends string>(array: T[], value: string): value is T => {
  // Downcast to less specific type, which is safe
  return (array as string[]).includes(value);
};

const gen3StaticMethods = ["Static1", "Static4"] as const satisfies Gen3StaticMethod[];

if (isType(gen3StaticMethods, values.method)) {
  setLatestHitAdv?.(values.advance, values.method);
}
```

src/rngToolsUi/gen3/wild/wild3CalibCaughtMon.component.tsx
```diff
@@ -284,98 +226,26 @@ export const Wild3CalibCaughtMon = ({
...
-          values.advance.adv_after_painting === targetPaintingAdvs.after &&
-          values.method === targetMethod
-        ) {
-          return "Target Pokémon";
-        }
+  const onUpdateCalib =
+    setLatestHitAdv == null
+      ? undefined
+      : (values: CaughtMonResult) => {
+          setLatestHitAdv?.(values.advance, values.method as Gen3Method);
```
zaksabeast: Same as the other comment, let's remove the cast here.

src/rngToolsUi/gen3/static/static3Calib.tsx
```diff
@@ -0,0 +1,232 @@
...
+import { calculateTargetSetupResult } from "./calculateTargetSetupResult";
+import { Gen3StaticMethod } from "~/rngTools";
+import Instructions_calib_with_battle_video from "./instructions_calib_with_battle_video.mdx";
+import Instructions_calib_without_battle_video from "./instructions_calib_without_battle_video.mdx";
+import {
+  buildGen3CalibFields,
+  buildGen3CalibPreviousStepFields,
+} from "../pokemonRng/calib";
+import { Static3TargetSetupInput } from "./static3TargetSetupInput";
+import Instructions_calib_skip_setup from "../wild/instructions_calib_skip_setup.mdx";
```
zaksabeast: This is pulling from the wild file, not the static file.

src/rngToolsUi/gen3/static/static3Calib.tsx
```diff
@@ -0,0 +1,232 @@
+import React from "react";
+import { Flex, MultiTimer, Field, Button } from "~/components";
+import { FormFieldTable } from "~/components/formFieldTable";
+import { TargetSetup } from "./static3TargetSetupSearcher";
+import { Static3CalibCaughtMon } from "./static3CalibCaughtMon";
+import { Gen3Console, gen3ConsoleFpsMap } from "~/types/console";
+import { BattleVideoInfo } from "../battleVideo/battleVideo";
+import { AllOrNone } from "~/types";
+import { BattleVideoInfoInput } from "../wild/wild3CalibBattleVideoInfoInput";
```
zaksabeast: Should this component be moved to a shared location instead of importing from wild?

zaksabeast (changes requested):

guides/Gen 3/Emerald/Gen 3 Static.mdx
RainingChain: What do you think of Starter, Legend, and Fossil RNG ? In Platinum, Legend, Gift, and Fossil RNG is used.

src/rngToolsUi/gen3/pokemonRng/shared.tsx
RainingChain: The convention I use is that the code that is shared between static3XXX.tsx and wild3XXX.tsx is put in a file called pokemonRng/XXX.tsx.

Now, my issue is where to I put the code shared by static3.tsx and wild3.tsx ? I guess I can rename to shared.tsx.

src/rngToolsUi/gen3/static/static3Calib.tsx
RainingChain: targetSetup only changes if targetSetupProp is not provided. 
Using the term initialTargetSetup seems to wrongly indicate that the targetSetup can change. I prefer to keep it targetSetupProp. `fixedTargetSetup` could also work if you prefer.

src/rngToolsUi/gen3/static/static3Calib.tsx
RainingChain: I'd rather not. It's quite short already.

RainingChain: That sounds a good idea, but I'm not sure how to code it.

If I put `if (isEqual(targetSetup, targetSetupProp)) {`, then I get the error
`React Hook React.useEffect has a missing dependency: 'targetSetup'. Either include it or remove the dependency array`

But I don't want the effect to be called when targetSetup changes.

guides/Gen 3/Emerald/Gen 3 Static.mdx
zaksabeast: Yeah, that sounds great

src/rngToolsUi/gen3/pokemonRng/shared.tsx
zaksabeast: `shared` is fine for now. I think `state` or `aceState` could work well too.

src/rngToolsUi/gen3/static/static3Calib.tsx
zaksabeast: Ah, if we want the component to use the prop when it's provided and use local state when it's not, `targetSetup`/`targetSetupProp` is good.

src/rngToolsUi/gen3/static/static3Calib.tsx
zaksabeast: What are your thoughts on this instead of the two local states:
```tsx
const [localTargetSetup, setLocalTargetSetup] = React.useState<TargetSetup | null>(null);
const targetSetup = targetSetupProp ?? localTargetSetup;

React.useEffect(() => {
  if (targetSetupProp == null) {
    return;
  }

  setLocalTargetSetup(null);
}, [targetSetupProp]);
```

RainingChain: I applied what you proposed. @zaksabeast

PR #460, opened 2026-06-23
Wild3/Static3 general improvements + fix

RainingChain:
- Display ability name instead of First/Second in result table
- Reusable Emerald AceForm component
- Uniformisation of labels and tooltip for Target advance label. 
- Painting advances tooltips now include "equivalent advances without painting reseeding" which is useful for tool-assisted players.
- Improved the tooltip for RNG-manipulated lead PID in Wild3.
- Fixed a bug in Static3 causing "Changing SID with ACE" box names to not be displayed.
- Fixed a bug where result Pokémon would be displayed as non-shiny even when "Changing SID with ACE" is active.
- Fixed a bug where "equivalent advances without painting reseeding" could be greater than U32_MAX.


src/rngToolsUi/gen3/pokemonRng/labels.tsx
```diff
@@ -0,0 +1,122 @@
...
+export const minFramesBeforePaintingLabel = () => ({
+  label: "Min frames before painting",
+  tooltip:
+    "For advanced users. Those frames ensure there is enough time between booting the game and interacting with the painting.",
+});
+
+export const minAdvsAfterPaintingLabel = () => ({
+  label: "Min advances after painting",
+  tooltip:
+    "To ensure there is enough time between interacting with the painting, and creating a Battle Video.",
```
zaksabeast: Tiny nitpick: we should remove the comma after `painting`

```
painting and creating a Battle Video.
````

File: src/rngToolsUi/gen3/wild/wild3TargetSetupInput.tsx [hunk 2]
```diff
@@ -276,7 +276,7 @@ export const Wild3TargetSetupInputFields = ({
   });
 
   const equivalentInitialAdvs =
-    lcrng_distance(0, targetFrameBeforePainting) + targetAdvance;
+    (lcrng_distance(0, targetFrameBeforePainting) + targetAdvance) % 2 ** 32;
```
zaksabeast, line 279: We're adding `% 2 ** 32` in several places. Should we have a display utility that handles this or a modulo argument in `lcrng_distance`?

src/rngToolsUi/gen3/static/static3TargetSetupSearcher.tsx
```diff
@@ -78,7 +79,7 @@ const convertToTargetSetup = (
       after: pidPath.advs.adv_after_painting,
     },
     targetMethod: pidPath.method,
-    aceSid: null,
+    aceSid: tidForAceSid,
```
zaksabeast: It looks like we're storing the tid in the `aceSid` field. Is this correct?

zaksabeast (changes requested):
One tiny nitpick and two questions

src/rngToolsUi/gen3/wild/wild3TargetSetupInput.tsx
RainingChain: The overflow doesn't actually occur in lcrng_distance. It's when we add  `+ targetAdvance` that it overflows. I think it's overkill to have a utility that does `% 2 ** 32`.

PR #465, opened 2026-07-12
Optimize Gen 3 ACE guide

fiaskotto:


fiaskotto:
I noticed now that there's also italian translation. I'll commit it as soon as possible!

PR #466, opened 2026-07-13
Add pokeradar rings generation logic (Gen 4)

fiaskotto:


rng_tools/src/generators/gen4/pokeradar.rs
```diff
@@ -0,0 +1,291 @@
...
+    target_advance: u32,
+    chain_count: u16,
+    battle_result: BattleResult,
+    selected_shake: ShakeType,
+) -> SimulateAdvanceResult {
+    let mut rng = Pokerng::new(init_seed);
+
+    for _ in 0..target_advance {
+        rng.next();
+    }
```
zaksabeast: We can use `rng.jump(target_advace)` to do this. It's a more performant version of the same logic.

rng_tools/src/generators/gen4/pokeradar.rs
```diff
@@ -0,0 +1,291 @@
...
+    #[test]
+    fn test_patch_ring() {
+        let result = simulate_advance(123456, 0, 10, BattleResult::Win, ShakeType::Fast);
+
+        for (expected_ring, patch) in result.patches.iter().enumerate() {
+            assert_eq!(patch.ring, expected_ring as i32);
+        }
+    }
+    #[test]
+    fn test_real_game_case_001() {
```
zaksabeast: Not a required change, but it might be a bit easier to read the expected results if formatted like this:
```rs
#[test]
fn test_real_game_case_001() {
    let result = simulate_advance(0x010D029F, 350, 1, BattleResult::Catch, ShakeType::Slow);
    let expected = [
        Patch {
            ring: 0,
            gx: 7,
            gz: 0,
            continue_chain: true,
            is_shiny: false,
            shake_type: ShakeType::Slow,
        },
        Patch {
            ring: 1,
            gx: 1,
            gz: 5,
            continue_chain: true,
            is_shiny: false,
            shake_type: ShakeType::Slow,
        },
        Patch {
            ring: 2,
            gx: 2,
            gz: 3,
            continue_chain: false,
            is_shiny: false,
            shake_type: ShakeType::Fast,
        },
        Patch {
            ring: 3,
            gx: 5,
            gz: 5,
            continue_chain: false,
            is_shiny: false,
            shake_type: ShakeType::Fast,
        },
    ];

    assert_list_eq!(result.patches, expected);
}
```

You'd need to derive `PartialEq` for `Patch`, and add `use crate::assert_list_eq;` under `mod tests`.

 `assert_list_eq` is like `assert_eq`, but with a cleaner output for lists.

rng_tools/src/generators/gen4/pokeradar.rs
```diff
@@ -0,0 +1,291 @@
...
+) -> SimulateAdvanceResult {
+    let mut rng = Pokerng::new(init_seed);
+
+    for _ in 0..target_advance {
+        rng.next();
+    }
+
+    let mut positions: [u16; 4] = [0; 4];
+    for (ring, slot) in positions.iter_mut().enumerate() {
+        *slot = rng.get_rand_range(RING_TILE_COUNT[ring]);
```
zaksabeast: I believe the game is using `DpptLogic::max` in all pokeradar logic.

For a lot of values,`DpptLogic::max` will produce the same result as `get_rand_range`, but not all values.

Here's an example showing the difference:
```rs
fn get_rand_range(rand: u16, max: u16) -> u16 {
    ((rand as u32 * max as u32) / 65536) as u16
}

#[test]
fn this_passes() {
    let max = 2;
    let high = 0xaaab;

    let result = get_rand_range(high, max);
    let expected = DpptLogic::max(high, max);

    assert_eq!(result, expected);
}

#[test]
fn this_fails() {
    let max = 3;
    let high = 0xaaab;

    let result = get_rand_range(high, max);
    let expected = DpptLogic::max(high, max);

    assert_eq!(result, expected);
}
```

Is there a reason to think `DpptLogic::max` isn't used for pokeradar?

rng_tools/src/generators/gen4/pokeradar.rs
```diff
@@ -0,0 +1,291 @@
...
+            } else {
+                ShakeType::Fast
+            };
+        } else {
+            shake_type = selected_shake;
+            if chain_count > 0 {
+                let sh_val = rng.get_rand_range(s_rate);
+                is_shiny = sh_val == 0;
+            }
+        }
```
zaksabeast: It looks like we can simplify this to:
```rs
let continue_chain = chain_count > 0 && cc_val < rates[ring as usize];

let (shake_type, is_shiny) = match continue_chain {
    true => {
        let sh_val = DpptLogic::max(rng.rand(), s_rate);
        (selected_shake, sh_val == 0)
    }
    false => {
        let sh_val = DpptLogic::max(rng.rand(), 100);
        let shake_type = match sh_val < 50 {
            true => ShakeType::Slow,
            false => ShakeType::Fast,
        };
        (shake_type, false)
    }
};
```

zaksabeast (changes requested):
Great job! This looks good.

I mostly have a few small cleanliness suggestions, and one small logic difference from the game.

rng_tools/src/generators/gen4/pokeradar.rs
fiaskotto: No, there's no valid reason, I just hadn't thought of it. Now I've configured everything to DpptLogic::max (even positions and cc_val), and test_real_game_case still passes with the same expected values.

PR #468, opened 2026-07-22
Refactor PokeRadar (Gen 4)

fiaskotto:


rng_tools/src/generators/gen4/pokeradar/simulate.rs
```diff
@@ -0,0 +1,134 @@
...
+use crate::rng::{Rng, lcrng::Pokerng};
+use wasm_bindgen::prelude::*;
+
+const RATES_WIN: [u16; 4] = [88, 68, 48, 28];
+const RATES_CATCH: [u16; 4] = [98, 78, 58, 38];
+
+#[wasm_bindgen]
+pub fn simulate_advance(
+    init_seed: u32,
+    target_advance: u32,
```
zaksabeast: Let's make `target_advance` a `usize` since the only place we use it casts to a `usize.`

rng_tools/src/generators/gen4/pokeradar/simulate.rs
```diff
@@ -0,0 +1,134 @@
...
+    chain_count: u16,
+    battle_result: BattleResult,
+    selected_shake: ShakeType,
```
zaksabeast: Let's move the parameters into a struct, like `PokeRadarAdvanceOpts`, `PokeRadar4AdvanceOpts`, or something similar.

There are a few benefits to this:

First, there's a rust lint rule for too many parameters, so if we add more in the future we'll need to use a struct anyways.

Second, it's easier to tell what named parameters are in TS:
```ts
simulate_advance(0xabcd, 10, 10, 'Catch', 'Fast')
```
vs
```ts
simulate_advance({
  init_seed: 0xabcd,
  target_advance: 10,
  chain_count: 10,
  battle_result: 'Catch',
  selected_shake: 'Fast'
})
```

Lastly, some of our utilities expect a single parameter to make portability easier.

rng_tools/src/generators/gen4/pokeradar/simulate.rs
```diff
@@ -0,0 +1,134 @@
...
+use super::types::{BattleResult, Patch, ShakeType, SimulateAdvanceResult};
+use crate::gen4::game_logic::{DpptLogic, GameSpecificLogic};
+use crate::rng::{Rng, lcrng::Pokerng};
+use wasm_bindgen::prelude::*;
+
+const RATES_WIN: [u16; 4] = [88, 68, 48, 28];
+const RATES_CATCH: [u16; 4] = [98, 78, 58, 38];
+
+#[wasm_bindgen]
+pub fn simulate_advance(
```
zaksabeast: Let's rename this to have `pokradar`, `pokeradar4`, or something uniquely identifying somewhere in the name.

Since all wasm functions on the TS side come from `~/rngTools`, it's hard to tell what game functions belong to without prefixing them.

Everything coming from `~/rngTools` also means we can't have two wasm functions with the same name. Prefixing allows us to have two `*_simulate_advance` functions in the future if we want.

rng_tools/src/generators/gen4/pokeradar/grid.rs
```diff
@@ -0,0 +1,163 @@
...
+pub const RING_TILE_COUNT: [u16; 4] = [32, 24, 16, 8];
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Tsify, Serialize, Deserialize)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub struct GridCoords {
+    pub x: i32,
+    pub z: i32,
+}
+
+#[wasm_bindgen]
```
zaksabeast: Is TS going to use this, or just the `simulate` function?

If TS won't use this, let's remove `wasm_bindgen` here.

rng_tools/src/generators/gen4/pokeradar/grid.rs
```diff
@@ -0,0 +1,163 @@
...
+    }
+    let r = rand - size * 2;
+    let ox = if r % 2 == 0 { 0 } else { size - 1 };
+    GridCoords {
+        x: ox,
+        z: r / 2 + 1,
+    }
+}
+
+#[wasm_bindgen]
```
zaksabeast: Similarly, will TS use this function?

zaksabeast (changes requested):
Just a few questions and some small tweaks, otherwise this looks good!

rng_tools/src/generators/gen4/pokeradar/grid.rs
fiaskotto: I'm still thinking about how to do the graphics, but I think I'll take inspiration from the existing Gen 6.. No, TS doesn't call these directly. It reads gx/gz already computed inside each Patch from pokeradar4_simulate_advance. Also removed wasm_bindgen and made these pub(crate).

zaksabeast (approved):

PR #469, opened 2026-07-24
Add step calculator (Gen 4)

fiaskotto:


rng_tools/src/generators/gen4/step_calculator.rs
```diff
@@ -0,0 +1,288 @@
...
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Tsify, Serialize, Deserialize)]
+#[tsify(into_wasm_abi, from_wasm_abi)]
+pub struct StepSearchOpts {
+    pub seed: u32,
+    pub initial_advance: usize,
+    pub current_cooldown_steps: u32,
+    /// True for walking/surfing, false for biking/tall grass.
+    pub is_walking: bool,
```
zaksabeast: I like how you used enums for the other fields!

I can see a future where we use this function in other tools, and someone accidentally sets `is_walking: true` for tall grass.

If we use an enum here like your other fields, that will make it easier to use correctly:
```rs
/// Movement type for the encounter search.
#[derive(Debug, Clone, Copy, PartialEq, Eq, Tsify, Serialize, Deserialize)]
#[tsify(into_wasm_abi, from_wasm_abi)]
pub enum EncounterMovement {
    Walking,
    Surfing,
    Biking,
    TallGrass,
}

impl EncounterMovement {
    fn base_rate(self) -> u32 {
        match self {
            EncounterMovement::Walking => 40,
            EncounterMovement::Surfing => 40,
            EncounterMovement::Biking => 70,
            EncounterMovement::TallGrass => 70,
        }
    }
}
```

rng_tools/src/generators/gen4/step_calculator.rs
```diff
@@ -0,0 +1,288 @@
...
+            ability: EncounterAbilityEffect::Half,
+            flute: EncounterFlute::Black,
+            item: EncounterItem::CleanseOrIncense,
+        };
+
+        let result = search_steps_to_encounter(opts).unwrap();
+
+        assert_eq!(result.total_steps, 92);
+        assert_eq!(result.encounter_rate, 1);
+        assert_eq!(result.seed_after, 0x026B57FA);
```
zaksabeast: Let's use the full struct here to guarantee we don't accidentally add a bug specific to the `missing_steps`, `cooldown_steps` values in the future:
```rs
let result = search_steps_to_encounter(opts).unwrap();
let expected = StepSearchResult { /* data */ };

assert_eq!(result, expected);
```

Using a struct also means any new fields we add in the future will give us type errors in all tests where we should test those new fields as well.

zaksabeast (changes requested):
Looks good!

Just two small tweaks to make sure I won't accidentally misuse the function in the future if I borrow it for other tools.

rng_tools/src/generators/gen4/step_calculator.rs
fiaskotto: I actually didn't put them on purpose since:

`total_steps = missing_steps + cooldown_steps`

I mean, if `total_step` is correct, they also has to be correct. But if you think they have to be there, we can put them.

zaksabeast (approved):

PR #470, opened 2026-07-27
Patch searcher

fiaskotto:
I added two tests, tell me if it's enough or I need to add more.


zaksabeast (changes requested):
Looks good! But our formatter check before deploying will fail (`bun run lint:rust:format`).

Can you run `cargo fmt`?

zaksabeast (approved):

PR #474, opened 2026-08-18
Add select button

fiaskotto:


zaksabeast (changes requested):
Please run `bun run format`, otherwise it looks good!

PR #479, opened 2026-09-01
Improve Emerald Shiny Starter Guide

RainingChain:
Key changes:
- Added battery-aware Emerald TID/SID and shiny-starter calculations in the Rust RNG tools.
- Added separate earliest valid advances and ideal TID/SID timing for dead-battery saves.
- Reworked the guide into a two-step interactive workflow using Stepper.
- Connected Step 1 results directly to Step 2, removing the need to re-enter TID and SID values.
- Added SID selection and cross-out controls for tracking tested candidates.
- Improved timer labels, calibration fields, offsets, human-input-delay controls, recommendations, and estimated completion times.

image
image
image
image


src/rngTools/index.ts
```diff
@@ -247,9 +247,10 @@ export const rngTools = new Proxy(
       ) => {
         const tools = await getRngTools();
         const func = tools[functionName];
-        // Casting as unknown[] fixes complex union errors.
         // The type signature guarantees this is used correctly.
-        return func(...(args as unknown[]));
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore -- Distributed union type from comlink makes this complex to type correctly. Sometimes, it's considered an error and sometimes not.
+        return func(...args);
```
zaksabeast: What's the reason for this change?

If it's sometimes considered an error, shouldn't we keep the existing types?

src/rngToolsUi/gen3/shinyStarter/generateTidSid.tsx
```diff
@@ -1,38 +1,40 @@
...
 export type FormState = z.infer<typeof Validator>;
 
+const OFFSET = 50;
+
 const initialValues: FormState = {
-  offset: 50, // offset on Emerald on mgba 0.11-8764
   tid: 0,
 };
 
 type Props = {
```
RainingChain: Another thing I want to add is the console type.

src/rngToolsUi/gen3/shinyStarter/targetPokemon.tsx
```diff
@@ -22,29 +22,39 @@ export const TargetPokemon = ({
...
+      if (!cancelled) {
+        setTargetStarter({ species, minMaxStats });
+        setTargetPokemonDesc(desc);
+      }
+    });
+
+    return () => {
+      cancelled = true;
+    };
+  }, [game, species, setTargetStarter, targetAdvance]);
```
zaksabeast, line 45: How come this was pulled out as a `useEffect`?

Was this handling a race condition if a user selected between different starters?

src/rngToolsUi/gen3/shinyStarter/index.tsx
```diff
@@ -19,51 +37,200 @@ type Props = {
...
+  );
+  const crossOutSid = (sid: number) => {
+    setTidSidResults((results) =>
+      results.map((result) =>
+        result.sid === sid ? { ...result, crossedOut: true } : result,
+      ),
+    );
+  };
+
+  React.useEffect(() => {
```
zaksabeast, line 147: Same here – is there a reason this is a useEffect instead of a callback when the user selects the target?

src/rngToolsUi/gen3/shinyStarter/generateTidSid.tsx
```diff
@@ -1,38 +1,47 @@
...
-  game: Game;
-};
-
-type Result = Gen3NearbySid & {
-  tid_gen_target_adv: number;
+  game: "emerald"; // The component and rust code only support emerald. RS is not supported.
```
zaksabeast: We might want to use a looser `string` since mdx doesn't has type validation.

We can validate `emerald` is the value in the component or return `null`.

Either that, or we can make assume these components are emerald exclusive and ignore the prop.

src/rngTools/index.ts
zaksabeast, line 253: What's the reason for this change?

If it's sometimes considered an error, shouldn't we keep the existing types?

zaksabeast (commented):
Mostly some questions

src/rngToolsUi/gen3/shinyStarter/targetPokemon.tsx
RainingChain: This fixes a bug. Without useEffect, when the component is shown the first time, the description and stats are 0 until the user selects something.
image

With the PR, Mudkip and valid stats are shown.

src/rngToolsUi/gen3/shinyStarter/index.tsx
RainingChain: Same

src/rngTools/index.ts
RainingChain: Like what my comment states, "Sometimes, it's considered an error and sometimes not."
Intellisense in vscode flags it as a an error. But `bun run lint` doesn't flag it as an error. `bun run check` does flag it as error. 
I tried to understand, but I gave up. It was getting in my way more than what the actual gain.

I can revert that part if you want.

zaksabeast: Oh strange, `bun run lint`, `bun run check`, and VSCode Intellisense haven't given me any errors with the existing cast.

I read your comment as the current line without `as unknown[]` sometimes errors. Since I hadn't seen errors with `as unknown[]`, it read like the cast had been removed, errors were observed, then `@ts-ignore` was added.

But it makes sense knowing `as unknown[]` wasn't working for you. I'm fine with keeping this as is.

src/rngToolsUi/gen3/shinyStarter/targetPokemon.tsx
zaksabeast: Nice, thanks – just wanted to check.

